PR #1, opened 2022-10-05
Zadachi

georgi-m-iliev:


Pract. 01/zad2.cpp
```diff
@@ -0,0 +1,13 @@
+#include <iostream>
+
+using namespace std;
+
+int main() {
+    int a = 5.4;
+    int b = 7.9;
```
dejansvetoslavov, line 7: think about the data type of the variables

Pract. 01/zad2.cpp
```diff
@@ -0,0 +1,13 @@
+#include <iostream>
+
+using namespace std;
+
+int main() {
+    double a = 5.4;
+    double b = 7.9;
+
+    cout << "P = " << 2 * a + 2 * b << endl;
```
Justsvetoslavov, line 9: Same as in mathematics:  2 * a + 2 * b <=> 2 * (a+b).
If it does not affect readability, we always should aim for a shorter code.

PR #15, opened 2022-10-12
Task 1

momouzunov2003:


SI/Sem.02/Pract.02/Task01.cpp
```diff
@@ -0,0 +1,21 @@
...
+//
+
+#include <iostream>
+using namespace std;
+
+int main()
+{
+    int year;
+    cin >> year;
+    if (year % 4 == 0 && year%100!=0 && year%400==0)
```
dejansvetoslavov, line 11: Add spaces between the operators

dejansvetoslavov (changes requested):

GeorgiTerziev02:
Solutions already uploaded

PR #16, opened 2022-10-12
task10_solution.cpp

milicavilica:
:)

### DIFF
--- a/SI/Sem.01/Pract.01/task10.cpp
+++ b/SI/Sem.01/Pract.01/task10.cpp
@@ -4,7 +4,7 @@ using namespace std;
 
 int main()
 {
-	int a, b;
+/*	int a, b;
 	cout << "a: ";
 	cin >> a;
 	cout << "b: ";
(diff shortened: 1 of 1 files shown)


SI/Sem.01/Pract.01/task10.cpp
```diff
@@ -17,5 +17,22 @@ int main()
...
+    double ak, bk, ck;
+    cout << "a = ";
+    cin >> ak;
+    cout << "b = ";
+    cin >> bk;
+    cout << "c = ";
+    cin >> ck;
+
+    double x1, x2;
+    x2 = ( - bk + sqrt(bk * bk - 4 * ak * ck)) / 2 * ak;
```
GeorgiTerziev02, line 32: Check if the discriminant is lower than zero

GeorgiTerziev02 (changes requested):

SI/Sem.01/Pract.01/task10.cpp
GeorgiTerziev02: My bad this is PR for Pract 01 - no ifs allowed, input is expected to be valid

GeorgiTerziev02:
We have already uploaded the solutions for week 1 [Here](https://github.com/Justsvetoslavov/Introduction_to_programming/tree/master/SI/Sem.01/Pract.01)

PR #19, opened 2022-10-12
task 1 and 2

milicavilica:


SI/Sem.02/Pract.02/ConsoleApplication_prakt2_12.10.22.cpp
```diff
@@ -0,0 +1,22 @@
...
+
+#include <iostream>
+
+using namespace std;
+
+int main()
+{
+    int year;
+    cin >> year;
+    if (year > 0 && year % 100 != 0 && year % 4 == 0)
```
dejansvetoslavov, line 12: The condition is not correct 

` For example, in the [Gregorian calendar](https://en.wikipedia.org/wiki/Gregorian_calendar), each leap year has 366 days instead of 365, by [extending February to 29 days](https://en.wikipedia.org/wiki/February_29) rather than the common 28. These extra days occur in each year that is an [integer](https://en.wikipedia.org/wiki/Integer) multiple of 4 (except for years evenly divisible by 100, but not by 400)`

SI/Sem.02/Pract.02/ConsoleApplication2.2.cpp
```diff
@@ -0,0 +1,40 @@
...
+        case 'O':
+        case 'u':
+        case 'U':
+        case 'y':
+        case 'Y':
+            cout << "The letter is a syllable." << endl;
+            break;
+        default:
+            cout << "The letter is not a syllable." << endl;
+            break;
```
dejansvetoslavov, line 33: The case blocks should be indented.

SI/Sem.02/Pract.02/ConsoleApplication2.2.cpp
```diff
@@ -0,0 +1,40 @@
...
+        }
+    }
+    else cout << "The character is not a latin letter." << endl;
```
dejansvetoslavov, line 36: Use {} and cout should be on new line

dejansvetoslavov:
The files name should be TaskXX.cpp where XX is the number of the task.

dejansvetoslavov (changes requested):

Justsvetoslavov:
This PR will be discussed in person on 19.10.
Topics:
- How to edit our PR.
- Commit separately.
- Commit messages.
- Description and order.

Justsvetoslavov:
Closing PR.

PR #21, opened 2022-10-12
Added solution for Task01

DrHristov:


dejansvetoslavov (approved):

SI/Sem.02/Pract.02/Task01.cpp
```diff
@@ -0,0 +1,17 @@
+#include <iostream>
+
+int main() {
+	int year = 0;
```
Justsvetoslavov, line 4: Is it me or do we have x2 tabulations ?

SI/Sem.02/Pract.02/Task01.cpp
```diff
@@ -0,0 +1,17 @@
...
+
+	std::cout << "Enter a year: ";
+	std::cin >> year;
+
+	if ((year % 400 == 0) || (year % 4 == 0 && year % 100 != 0)) {
+		std::cout << "Entered year is leap.\n";
+	}
```
Justsvetoslavov, line 11: Suggestion: 
} else {

SI/Sem.02/Pract.02/Task01.cpp
DrHristov: It's not just you, I can see them as well. it is definitely not on purpose. There should be only one tabulation. I don't know why that has happened, but in the source file the tabs are single. I've attached a screenshot of the source file that I've uploaded here.
Task01Screenshot

SI/Sem.02/Pract.02/Task01.cpp
DrHristov: I completely agree. Changed the setting in VS to automatically put it there. From now on the "else" will be there :D

SI/Sem.02/Pract.02/Task01.cpp
DrHristov: I resolved the issue. It's just GitHub that displays the single tab so big that it looks like 2 tabs. But in reality, when copying the code from the file and pasting it in VS the tab is only 1. In conclusion it's just a visual "bug".

PR #22, opened 2022-10-15
УП-практикум_12.10.22г.

kokopeevski:
Калоян Кирилов Пеевски ( II група ) - факултетен номер: 2MI0600170

### DIFF
--- a/SI/Sem.02/Pract.02/Task10.cpp
+++ b/SI/Sem.02/Pract.02/Task10.cpp
@@ -1,0 +1,52 @@
+﻿// Task10.cpp : This file contains the 'main' function. Program execution begins and ends there.
+//
+
+#include <iostream>
+using namespace std;
+
+int main()
+{
+	double a, b;
+	char operation;
+	cout << "Enter the first number: ";
+	cin >> a;
+	cout << endl;
+
+	cout << "Choose possible operation from (+, -, *, /): ";
+	cin >> operation;
+	cout << endl;
+
+	cout << "Enter the second number: ";
+	cin >> b;
+	cout << endl;
+
+	double res = 0;
+	switch (operation)
+	{
+		case '+':
+			res = a + b;
+			break;
+		case '-':
+			res = a - b;
+			break;
+		case '*':
+			res = a * b;
+			break;
+		case '/':
+			if (b == 0)
+			{
+				cout << "The divisor cannot be 0" << endl;
+				break;
+			}
+			else
+			{
+				res = a / b;
+				break;
+			}
+		default:
+			cout << "Invalid operation" << endl;
+			break;
+	}
+	if (res != 0) cout << "Result: " << res << endl;
+}
+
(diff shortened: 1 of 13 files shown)


SI/Sem.02/Pract.02/Task9.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+#include <iostream>
+using namespace std;
+
+int main()
+{
+	int firstNum, secondNum, thirdNum;
+	cout << "Enter three numbers: ";
+	cin >> firstNum >> secondNum >> thirdNum;
+	if (firstNum < secondNum && firstNum < thirdNum && secondNum < thirdNum)
```
GeorgiTerziev02, line 12: This solution doesn't work for the input 5 6 5 and there is a much simpler solution with just reordering(swapping) the numbers with three ifs

SI/Sem.02/Pract.02/Task10.cpp
```diff
@@ -0,0 +1,52 @@
...
+			else
+			{
+				res = a / b;
+				break;
+			}
+		default:
+			cout << "Invalid operation" << endl;
+			break;
+	}
+	if (res != 0) cout << "Result: " << res << endl;
```
GeorgiTerziev02, line 50: This solution doesn't give output if your input is 5,+,-5 (the sum of 5 and -5 is zero)

SI/Sem.02/Pract.02/Task11.cpp
```diff
@@ -0,0 +1,56 @@
...
+#include <iostream>
+using namespace std;
+
+int main()
+{
+	int a, b, c;
+	cout << "Enter three numbers: ";
+	cin >> a >> b >> c;
+	int max, mid, min;
+	if (a>=b && a>=c)
```
GeorgiTerziev02, line 13: Your solution works but as similar to task 9 you can swap the numbers between them with three ifs and then print a*100 + b * 10 + c

SI/Sem.02/Pract.02/Task12.cpp
```diff
@@ -0,0 +1,53 @@
...
+using namespace std;
+
+int main()
+{
+	int day, month;
+	cout << "Enter a date(day month): ";
+	cin >> day >> month;
+	cout << endl;
+	
+	if (month == 1 || month == 3 || month == 5 || month == 7 || month == 8 || month == 10 || month == 12)
```
GeorgiTerziev02, line 14: The solution doesn't handle the case 31 12

GeorgiTerziev02 (changes requested):
There are some mistakes and missed edge cases but overally looks good.

SI/Sem.02/Pract.02/Task3.cpp
```diff
@@ -0,0 +1,15 @@
...
+#include <iostream>
+using namespace std;
+
+int main()
+{
+	int a, b;
+	cout << "Enter a two numbers: ";
+	cin >> a >> b;
+
+	(a > b) ? (cout << "The first one is bigger") : (cout << "The second one is bigger");
```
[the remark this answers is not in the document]
Justsvetoslavov: Should always avoid repetition.

Justsvetoslavov:
Review the new written PR rules and open new PRs if you would like (also check which ones are already merged ,as they will be duplicates).

PR #27, opened 2022-10-16
Prakt. 2

stefka04:
Stefka Lyaskalieva - 4MI0600158

### DIFF
--- a/SI/Sem.02/Pract.02/Task2.cpp
+++ b/SI/Sem.02/Pract.02/Task2.cpp
@@ -1,0 +1,33 @@
+// ConsoleApplicationPrakt2Task2.cpp : This file contains the 'main' function. Program execution begins and ends there.
+//
+
+#include <iostream>
+using namespace std;
+int main()
+{
+    char symbol;
+    cout << "Enter a symbol";
+    cin >> symbol;
+    if ('A' <= symbol && symbol<= 'Z' || 'a' <= symbol && symbol <= 'z')
+    {
+        switch (symbol)
+        {
+        case 'a': ;
+        case 'e':;
+        case 'i':;
+        case 'o':;
+        case 'u':;
+        case 'y':;
+        case 'A':;
+        case 'E':;
+        case 'I':;
+        case 'O':;
+        case 'U':;
+        case 'Y': cout << "The symbol is a vowel letter"; break;
+        default: cout << "The symbol is a consonant letter";
+
+        }
+    }
+
+}
+
--- a/SI/Sem.02/Pract.02/Task2.cpp
+++ b/SI/Sem.02/Pract.02/Task2.cpp
@@ -1,33 +1,31 @@
-// ConsoleApplicationPrakt2Task2.cpp : This file contains the 'main' function. Program execution begins and ends there.
-//
-
 #include <iostream>
 using namespace std;
 int main()
 {
     char symbol;
-    cout << "Enter a symbol";
+    cout << "Enter a symbol: ";
     cin >> symbol;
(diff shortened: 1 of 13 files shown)


SI/Sem.02/Pract.02/Task2.cpp
```diff
@@ -0,0 +1,33 @@
...
+int main()
+{
+    char symbol;
+    cout << "Enter a symbol";
+    cin >> symbol;
+    if ('A' <= symbol && symbol<= 'Z' || 'a' <= symbol && symbol <= 'z')
+    {
+        switch (symbol)
+        {
+        case 'a': ;
```
plamena-ilieva, line 15: When you declare cases like that, you do not need semicolon (;) - it is an empty statement.

SI/Sem.02/Pract.02/Task3.cpp
```diff
@@ -0,0 +1,14 @@
...
+//
+
+#include <iostream>
+using namespace std;
+int main()
+{
+	int number1, number2;
+	cout << "Enter two numbers"<<endl;
+	cin >> number1>> number2;
+	cout << ((number1 != number2 && number1 > number2) ? "The first one is larger" : "The second one is smaller");
```
plamena-ilieva, line 11: If number1 > number2, they are never equal, so you do not need to check number1 != number2. "The second one is smaller" will be printed if the condition is false, which means number1 <= number2, but in this case the second is bigger. Also, you need to make sure the numbers are different in both cases.

SI/Sem.02/Pract.02/Task4.cpp
```diff
@@ -0,0 +1,39 @@
...
+int main()
+{
+    int grade;
+    cout << "Result(%): ";
+    cin>> grade;
+    if (grade >= 90)
+    {
+        cout << "Grade: 6+";
+    }
+    if (grade >= 80 && grade<90)
```
plamena-ilieva, line 15: The program works correctly, if the number is within range [0, 100] but even if you find the condition you need, you will also check all the other conditions, which can be avoided if you use _else if_. That will spare you some of the extra boolean expressions, too.

SI/Sem.02/Pract.02/Task6.cpp
```diff
@@ -0,0 +1,34 @@
...
+int main()
+{
+	double a, b, c, x1, x2, D;
+	cout << "a = ";
+	cin >> a;
+	cout << "b = ";
+	cin >> b;
+	cout << "c = ";
+	cin >> c;
+	D = sqrt(b * b - 4 * (a * c));
```
plamena-ilieva, line 14: D = b * b - 4 * (a * c);
You have to check whether D < 0 before you use sqrt, because you can only calculate square root of non negative numbers. Then, after the checking, you can calculate sqrt(D).

SI/Sem.02/Pract.02/Task7.cpp
```diff
@@ -0,0 +1,27 @@
...
+
+#include <iostream>
+using namespace std;
+
+int main()
+{
+    char symbol;
+    cout << "Enter a symbol: ";
+    cin >> symbol;
+    if (symbol >= 65 && symbol <= 90)
```
plamena-ilieva, line 11: The program works correctly. However, if you have to search each symbol's code every time, it takes time. When you have an exam, you will not be allowed to use the Internet to check the ASCII table and you do not need to learn it by heart. My advice is to use the symbols themselves:
if (symbol >= 'A' && symbol <= 'Z')
symbol -= 'a'-'A';

Justsvetoslavov:
Closing PR in terms of duplicate.
The tasks are not in separate commits and we cannot c**herry-pick solutions.**

Review the new written PR rules.

PR #28, opened 2022-10-17
Домашни практикум

georgibaltiev:
Георги Балтиев, СИ курс 1 група 2, 2MI0600168


SI/Sem.02/Pract.02/task2.cpp
```diff
@@ -0,0 +1,29 @@
+#include <iostream>
+using namespace std;
+int main() {
+    char a;
+    cout<<"Please enter a symbol:"<<endl;
+    cin>>a;
+    bool isVowel;
+    if((a >= 65 && a <= 90) || (a >= 97 && a <= 122)) {
```
plamena-ilieva, line 8: The program works correctly. However, if you have to search each symbol's code every time, it takes time. When you have an exam, you will not be allowed to use the Internet to check the ASCII table and you do not need to learn it by heart. My advice is to use the symbols themselves:
if (a >= 'A' && a <= 'Z')

SI/Sem.02/Pract.02/task7.cpp
```diff
@@ -0,0 +1,12 @@
+#include <iostream>
+using namespace std;
+int main() {
+    char a;
+    cout<<"Please enter a letter: "<<endl;
+    cin>>a;
+    if((a >= 65 && a <= 90) || (a >= 97 && a <= 122)) {
+        if(a >= 65 && a <= 90) a += 32;
```
plamena-ilieva, line 8: Same as the previous comment,
a += 'a'-'A';

Justsvetoslavov:
Pract.01 solutions are added.
Separate Pract.02 solutions if you want them to be merged in separate PRs.
Closing PR with label **dupilcate**.

PR #31, opened 2022-10-19
Pract 2

milicavilica:


Justsvetoslavov:
Review the new written PR rules and open new PRs if you would like (also check which ones are already merged ,as they will be duplicates).
For example Pract.01 solutions are added.

PR #32, opened 2022-10-19
Add files via upload

kmtoncheva:


Justsvetoslavov:
Review the new written PR rules which you can view  [here](https://github.com/Justsvetoslavov/Introduction_to_programming).
Rename the files & open new PR/s. After that please include me or tag me for review.
If you are having any problems, ping me.

GeorgiTerziev02:
For pract 02 the new rules for the PRs will not be considered. Do not close this pull request.

Justsvetoslavov:
@GeorgiTerziev02 We cannot cherry-pick due to single commits.
Do you still want to keep the PR ?

SI/Sem.02/Pract.02/task_1_p02.cpp
```diff
@@ -0,0 +1,19 @@
+
+#include <iostream>
+
+int main()
+{
+    int year;
+    std::cout << "Enter a year: ";
+    std::cin >> year;
+    if ((0 == year % 4 )||( 0 != year % 100 && 0 == year % 400) )
```
GeorgiTerziev02, line 9: year % 4 == 0 is more readable way of writing code
Because you first write the expression and then you write what you expect to be equal to

SI/Sem.02/Pract.02/task_4_p02.cpp
```diff
@@ -0,0 +1,43 @@
...
+        {
+            std::cout << "Grade : 5";
+        }
+        if (result >= 80 && result < 90)
+        {
+            std::cout << "Grade : 6";
+        }
+        if (result >= 90)
+        {
+            std::cout << "Grade : 6 + ";
```
GeorgiTerziev02, line 33: This should be an else if construction
With else if you will prevent making useless checks and
        if (result < 40) {
            std::cout << "Grade : 2";
        }
        if (result >= 40 && result < 60) {
            std::cout << "Grade : 3";
        }
and here you you won't have to write result >= 40 in the second if

SI/Sem.02/Pract.02/task_6_p02.cpp
```diff
@@ -0,0 +1,34 @@
...
+    std::cout << "c = ";
+    std::cin >> c;
+    double D = b*b - 4*a*c;
+    if (D > 0)
+    {
+        double X1 = (-b - sqrt(D)) / 2 * a;
+        double X2 = (-b + sqrt(D)) / 2 * a;
+        std::cout << "X1 = " << X1 << std::endl << "X2 = " << X2;
+    }
+    if (D == 0)
```
GeorgiTerziev02, line 23: else if

SI/Sem.02/Pract.02/task_7_p02.cpp
```diff
@@ -0,0 +1,30 @@
...
+
+#include <iostream>
+
+int main()
+{
+    char symbol;
+    std::cout << "Enter a letter: ";
+    std::cin >> symbol;
+    char LowerToUpper = symbol - 32;
+    char UpperToLower = symbol + 32;
```
GeorgiTerziev02, line 11: You should not do these operations in advance when you will need only one of them
You should do these operations when you enter the if for LowerToUpper
Also, use camelCase for naming you variables

SI/Sem.02/Pract.02/task_8_p02.cpp
```diff
@@ -0,0 +1,33 @@
+
+
+#include <iostream>
+
+int main()
+{
+    int number = 0;
```
GeorgiTerziev02, line 7: You don't need to intialize the variable with value, because you will overwrite it with the cin

SI/Sem.02/Pract.02/task_9_p02.cpp
```diff
@@ -0,0 +1,52 @@
...
+    int a = 0;
+    int b = 0;
+    int c = 0;
+    std::cout << "a= ";
+    std::cin >> a;
+    std::cout << "b= ";
+    std::cin >> b;
+    std::cout << "c= ";
+    std::cin >> c;
+    if (a < b && a<c)
```
GeorgiTerziev02, line 16: Think of a better way to do this task
Hint: swap the numbers between them if some condition is true

SI/Sem.02/Pract.02/task_10_p02.cpp
```diff
@@ -0,0 +1,29 @@
...
+
+int main()
+{
+    int a = 0;
+    int b = 0;
+    std::cout << "a = ";
+    std::cin >> a;
+    std::cout << "b = ";
+    std::cin >> b;
+    int SUM = a + b;
```
GeorgiTerziev02, line 13: use camelCase for naming you variables

SI/Sem.02/Pract.02/task_10_p02.cpp
```diff
@@ -0,0 +1,29 @@
+
+
+#include <iostream>
+
+int main()
+{
+    int a = 0;
+    int b = 0;
+    std::cout << "a = ";
+    std::cin >> a;
```
GeorgiTerziev02, line 10: The point of the task is to receive from the console a, b and character that is +,-,*,/ and based on the characted you should do a specific operation

SI/Sem.02/Pract.02/task_11_p02.cpp
```diff
@@ -0,0 +1,51 @@
...
+    int a = 0;
+    int b = 0;
+    int c = 0;
+    std::cout << "a = ";
+    std::cin >> a;
+    std::cout << "b = ";
+    std::cin >> b;
+    std::cout << "c = ";
+    std::cin >> c;
+    if (a > b && a > c && a != 0)
```
GeorgiTerziev02, line 16: Same problem and hint as in task 9

GeorgiTerziev02 (changes requested):

GeorgiTerziev02:
Code review done. Think on the comments.
Closing PR because it is not meeting the new PR rules.

PR #33, opened 2022-10-19
Pract02

Danail3:


SI/Sem.02/Solutions/Task03.cpp
```diff
@@ -0,0 +1,11 @@
+#include <iostream>
+using namespace std;
+
+int main()
+{
+	double number1, number2;
+	cin >> number1 >> number2;
+	cin.ignore();
```
Justsvetoslavov, line 8: Why do you need cin.ignore() ?

SI/Sem.02/Solutions/Task03.cpp
```diff
@@ -0,0 +1,11 @@
...
+	double finalResult = number1 > number2 ? number1 : number2;
```
Justsvetoslavov, line 9: If we are receiving 2 different numbers it is correct, however, can we check whether they are equal?

SI/Sem.02/Solutions/Task04.cpp
```diff
@@ -0,0 +1,35 @@
...
+	}
+	else if (result >= 60)
+	{
+		cout << "Grade: 4";
+	}
+	else if (result >= 40)
+	{
+		cout << "Grade: 3";
+	}
+	else
```
Justsvetoslavov, line 29: Can we have negative percentages?

SI/Sem.02/Solutions/Task05.cpp
```diff
@@ -0,0 +1,17 @@
+#include <iostream>
+
+using namespace std;
+
+int main()
+{
+	double sideA, sideB, sideC;
+	cin >> sideA >> sideB >> sideC;
+	if ((sideA + sideB) > sideC && (sideA + sideC) > sideB && (sideB + sideC) > sideA)
```
Justsvetoslavov, line 9: if ((a + b > c) && (a + c > b) && (b + c > a)),
would it be better to group the cases, rather than the addition?

SI/Sem.02/Solutions/Task05.cpp
```diff
@@ -0,0 +1,17 @@
...
+	{
+		cout << "It's a valid triangle";
```
Justsvetoslavov, line 11: New lines are always a good practice.
Use std::endl or "\n:"

SI/Sem.02/Solutions/Task06.cpp
```diff
@@ -0,0 +1,29 @@
+#include <iostream>
+#include <cmath>
+using namespace std;
+
+int main()
+{
+	double coefA, coefB, coefC;
```
Justsvetoslavov, line 7: coef is not necessary in this case in my opinion.

Also , try to have a more User-friendly input:
For example:
	 cout << "Enter the first number: ";
	 cin >> a;

SI/Sem.02/Solutions/Task06.cpp
```diff
@@ -0,0 +1,29 @@
...
+	else if(D == 0)
+	{
+		double root;
+		root =  ( - coefB ) / (2 * coefA);
+		cout << "The root of the equation is " << root << endl;
+	}
+	else
+	{
+		double root1, root2;
+		root1 = ((-coefB) + sqrt(D)) / 2 * coefA;
```
Justsvetoslavov, line 23: if a is 0, then this is not a quadratic equation!

SI/Sem.02/Solutions/Task06.cpp
```diff
@@ -0,0 +1,29 @@
+#include <iostream>
+#include <cmath>
+using namespace std;
+
+int main()
+{
+	double coefA, coefB, coefC;
+	cin >> coefA >> coefB >> coefC;
+	double D = coefB * coefB - 4 * coefA * coefC;
```
Justsvetoslavov, line 9: after we verify 'a' , we can create variables for roots and discriminant.

SI/Sem.02/Solutions/Task06.cpp
```diff
@@ -0,0 +1,29 @@
...
+	{
+		double root;
+		root =  ( - coefB ) / (2 * coefA);
+		cout << "The root of the equation is " << root << endl;
+	}
+	else
+	{
+		double root1, root2;
+		root1 = ((-coefB) + sqrt(D)) / 2 * coefA;
+		root2 = ((-coefB) - sqrt(D)) / 2 * coefA;
```
Justsvetoslavov, line 24: root2 = (-b - sqrt(discriminant)) / (2 * a);
Again the brackets are not required.

SI/Sem.02/Solutions/Task07.cpp
```diff
@@ -0,0 +1,23 @@
+#include <iostream>
+using namespace std;
+
+int main()
+{
+	char letter;
+	cin >> letter;
+	if (letter >= 65 && letter <= 90)
```
Justsvetoslavov, line 8: let us work with 'a' and 'z'.
In terms of functionality & performance -> the same.
In terms of readability, however, they are better.

SI/Sem.02/Solutions/Task07.cpp
```diff
@@ -0,0 +1,23 @@
...
+	{
+		letter += 32;
```
Justsvetoslavov, line 10: Let us create a const variable for this '32' and use it.

Justsvetoslavov (changes requested):
Currently reviewed until Task08(excluding it).

Justsvetoslavov:
Closing PR due to inactivity and task solutions will be published.

PR #37, opened 2022-10-19
Pract.03: Add 01-03 | Presiyana Borisova

PresiyanaBB:
Solutions for tasks from 1 to 3 from Prac03
from Presiyana - 3MI0600240


dejansvetoslavov:
Each file should be committed separately

PR #38, opened 2022-10-19
Pract.03: Add 01-03 | Danail Todorov

Danail3:
Danail Todorov
6 0MI0600241

### DIFF
--- a/SI/Sem.03/Pract.03/Task03.cpp
+++ b/SI/Sem.03/Pract.03/Task03.cpp
@@ -1,0 +1,20 @@
+#include <iostream>
+using namespace std;
+
+int main()
+{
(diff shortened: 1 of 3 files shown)


SI/Sem.03/Pract.03/Task03.cpp
```diff
@@ -0,0 +1,20 @@
...
+	{
+		number /= 10;
+	}
+	while (number > 0)
+	{
+		left = number % 10;
+		cout << left;
+		number /= 10;
+	}
+}
```
dejansvetoslavov, line 20: Try input 8

dejansvetoslavov (changes requested):

PR #40, opened 2022-10-19
Pract.02: Add 04-08 | Atanas Hristov

AtanasHrist0v:
Atanas Hristov
6 8MI0600242

### DIFF
--- a/SI/Sem.02/Pract.02/Task04.cpp
+++ b/SI/Sem.02/Pract.02/Task04.cpp
@@ -1,30 +1,30 @@
 #include <iostream>
 
 int main() {
-	int percentage = 0, grade = 2;
+	int percentage = 0, grade = 0;
 
 	std::cout << "Result: ";
 	std::cin >> percentage;
 
 	if (percentage >= 0 && percentage <= 100) {
-		if (percentage >= 40) {
+		bool isSixPlus = false;
+
+		if (percentage >= 90) {
+			grade = 6;
+			isSixPlus = true;
+		} else if (percentage >= 80) {
+			grade = 6;
+		} else if (percentage >= 70) {
+			grade = 5;
+		} else if (percentage >= 60) {
+			grade = 4;
+		} else if (percentage >= 40) {
 			grade = 3;
-			if (percentage >= 60) {
-				grade = 4;
-				if (percentage >= 70) {
-					grade = 5;
-					if (percentage >= 80) {
-						grade = 6;
-						if (percentage >= 90) {
-							std::cout << "Grade: " << grade << "+\n";
-							return 0;
-						}
-					}
-				}
-			}
+		} else {
+			grade = 2;
 		}
 
-		std::cout << "Grade: " << grade << std::endl;
+		std::cout << "Grade: " << grade << (isSixPlus ? "+\n" : "\n");
 	} else {
 		std::cout << "Entered percentage is not valid!\n";
 	}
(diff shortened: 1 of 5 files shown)


SI/Sem.02/Pract.02/Task04.cpp
```diff
@@ -0,0 +1,33 @@
...
+					if (percentage >= 80) {
+						grade = 6;
+						if (percentage >= 90) {
+							std::cout << "Grade: " << grade << "+\n";
+							return 0;
+						}
+					}
+				}
+			}
+		}
```
dejansvetoslavov, line 25: Refactor! It is really hard to read this block of code.

SI/Sem.02/Pract.02/Task05.cpp
```diff
@@ -0,0 +1,21 @@
...
+	double sideA = 0, sideB = 0, sideC = 0;
+
+	std::cout << "Enter value for side a: ";
+	std::cin >> sideA;
+	std::cout << "Enter value for side b: ";
+	std::cin >> sideB;
+	std::cout << "Enter value for side c: ";
+	std::cin >> sideC;
+
+	if ((sideA + sideB <= sideC) || (sideA + sideC <= sideB) || (sideB + sideC <= sideA)) {
```
dejansvetoslavov, line 13: Wrong conditions
https://www.wikihow.com/Determine-if-Three-Side-Lengths-Are-a-Triangle

SI/Sem.02/Pract.02/Task07.cpp
```diff
@@ -0,0 +1,19 @@
+#include <iostream>
+
+int main() {
+	char input;
+
+	std::cin >> input;
+
+	if (input >= 'A' && input <= 'Z') {
+		std::cout << char(int(input) + 32) << std::endl;
```
dejansvetoslavov, line 9: the casts to int and char are unneeded

SI/Sem.02/Pract.02/Task07.cpp
```diff
@@ -0,0 +1,19 @@
...
+	}
+	else if (input >= 'a' && input <= 'z') {
+		std::cout << char(int(input) - 32) << std::endl;
```
dejansvetoslavov, line 12: the casts to int and char are unneeded

SI/Sem.02/Pract.02/Task08.cpp
```diff
@@ -0,0 +1,53 @@
...
+			case 7:
+				std::cout << "VII";
+				break;
+			case 8:
+				std::cout << "VIII";
+				break;
+			case 9:
+				std::cout << "IX";
+				break;
+			case 10:
```
dejansvetoslavov, line 40: Is it possible to hit this case?

dejansvetoslavov (changes requested):

SI/Sem.02/Pract.02/Task05.cpp
AtanasHrist0v: They're not wrong, because the cout in the if says "... is invalid...". I've just used de Morgan's Theorem and flipped the conditions. But I will say that it's definitely more confusing this way so I'll update it regardless.

SI/Sem.02/Pract.02/Task07.cpp
AtanasHrist0v: If I remove both casts I end up with this code "std::cout << input + 32 << std::endl;" and if the input is "A" the output is 97, not the character a. So then maybe I do need at least the char cast?

Justsvetoslavov: Ping us tomorrow to discuss it.

Justsvetoslavov (approved):

PR #43, opened 2022-10-19
Pract03: Add 04-05 | Presiyana Borisova

PresiyanaBB:
Tasks 04 and 05 solution
Presiyana - 3MI0600240

### DIFF
--- a/SI/Sem.03/Pract.03/Task04.cpp
+++ b/SI/Sem.03/Pract.03/Task04.cpp
@@ -1,0 +1,23 @@
+#include <iostream>
+using namespace std;
+
+int main()
+{
+    int m;
+    int n;
+
+    cout << "Please enter 2 whole positive numbers: ";
+
+    cin >> m >> n;
+
+    for (int i = 1; i <= n*m; i++)
+    {
+        cout << i << " ";
+
+        if (i%n == 0)
+        {
+            cout << "\n";
+        }
+
+    }
+}
--- a/SI/Sem.03/Pract.03/Task05.cpp
+++ b/SI/Sem.03/Pract.03/Task05.cpp
@@ -1,0 +1,26 @@
+#include <iostream>;
+using namespace std;
+
+int main()
+{
+    int targetNumber;
+    int counter = 0;
+    int guessNumber;
+
+    cout << "Please enter whole number from 0 to 100: ";
+    cin >> targetNumber;
+
+    do
+    {
+        cout << "Enter your guess: ";
+
+        cin >> guessNumber;
+        counter++;
+
+        if (guessNumber == targetNumber)
+        {
+            cout << "You won with " << counter << " attempts";
+            break;
+        }
+    } while (true);
+}


SI/Sem.03/Pract.03/Task04.cpp
```diff
@@ -0,0 +1,23 @@
...
+
+    cout << "Please enter 2 whole positive numbers: ";
+
+    cin >> m >> n;
+
+    for (int i = 1; i <= n*m; i++)
+    {
+        cout << i << " ";
+
+        if (i%n == 0)
```
Justsvetoslavov, line 17: if (i % n == 0)

SI/Sem.03/Pract.03/Task05.cpp
```diff
@@ -0,0 +1,26 @@
...
+using namespace std;
+
+int main()
+{
+    int targetNumber;
+    int counter = 0;
+    int guessNumber;
+
+    cout << "Please enter whole number from 0 to 100: ";
+    cin >> targetNumber;
```
Justsvetoslavov, line 11: targetNumber is between 0 and 100 , add a check.

SI/Sem.03/Pract.03/Task05.cpp
```diff
@@ -0,0 +1,26 @@
+#include <iostream>;
+using namespace std;
+
+int main()
+{
+    int targetNumber;
+    int counter = 0;
```
Justsvetoslavov, line 7: Suggestion: move the int counter = 0; just before the loop.

SI/Sem.03/Pract.03/Task05.cpp
```diff
@@ -0,0 +1,26 @@
...
+
+        cin >> guessNumber;
+        counter++;
+
+        if (guessNumber == targetNumber)
+        {
+            cout << "You won with " << counter << " attempts";
+            break;
+        }
+    } while (true);
```
Justsvetoslavov, line 25: We are looping until the numbers are equal.
Can we avoid/change the while (true) ?

SI/Sem.03/Pract.03/Task05.cpp
```diff
@@ -0,0 +1,26 @@
...
+    cout << "Please enter whole number from 0 to 100: ";
+    cin >> targetNumber;
+
+    do
+    {
+        cout << "Enter your guess: ";
+
+        cin >> guessNumber;
+        counter++;
+
```
Justsvetoslavov, line 19: if (guessNumber < targetNumber)
				cout << "The number you are looking for is greater than your guess!\n";
 else if (guessNumber > targetNumber)
				cout << "The number you are looking for is less than your guess!\n";

SI/Sem.03/Pract.03/Task05.cpp
```diff
@@ -0,0 +1,26 @@
...
+        if (guessNumber == targetNumber)
+        {
+            cout << "You won with " << counter << " attempts";
```
Justsvetoslavov, line 22: Should be outside of the loop.
Our condition is in the while (<condition>)

Justsvetoslavov (changes requested):

GeorgiTerziev02:

PR #44, opened 2022-10-19
Prakt.03: Add 01-03| Margarita Gencheva

Margarita-Gencheva2003:


SI/Sem.03/Pract.03/Task02.cpp
```diff
@@ -0,0 +1,14 @@
...
+	int n = 1;
+	std::cin >> n;
+	for (int i = 1; i < n; i++) {
+		for (int j = n - 1; j > 0; j--) {
+			if (i + j == n) {
+				std::cout << "x = " << i << " y = " << j << std::endl;
+			}
+		}
+	}
+
```
dejansvetoslavov, line 13: Delete blank line

SI/Sem.03/Pract.03/Task02.cpp
```diff
@@ -0,0 +1,14 @@
+#include <iostream>
+
+int main() {
+	int n = 1;
+	std::cin >> n;
+	for (int i = 1; i < n; i++) {
+		for (int j = n - 1; j > 0; j--) {
```
dejansvetoslavov, line 7: Think about how you can solve the task without nested loop

Margarita-Gencheva2003: #include <iostream>
int main() {
	int n = 1, addend = 1;
	std::cin >> n;
	addend = n;
	for (int i = 1; i < n; i++) {
		addend--;
		std::cout << "x = " << i << " y = " << addend << std::endl;
	}
}

PR #49, opened 2022-10-19
Pract.03: Add 01-03 | Miroslav Stoyanovski

mir4oni3:
Miroslav Stoyanovski
6 5MI0600195

### DIFF
--- a/SI/Sem.03/Pract.03/Task04.cpp
+++ b/SI/Sem.03/Pract.03/Task04.cpp
@@ -1,0 +1,15 @@
+#include <iostream>
+using namespace std;
+int main()
+{
+	int m, n;
+	cin >> m >> n;
+	for (int i = 1; i <= m * n;) {
+		for (int j = 0; j < n; j++) {
+			cout << i << " ";
+			i++;
+		}
+		cout << endl;
+	}
+
+}
--- a/SI/Sem.03/Pract.03/Task05.cpp
+++ b/SI/Sem.03/Pract.03/Task05.cpp
@@ -1,0 +1,26 @@
+#include <iostream>
+using namespace std;
+int main()
+{
+	int n;
+	cout << "enter initial number" << endl;
+	cin >> n;
+	int cur = -1;
+	int guesses = 0;
+	while (true) {
+		cout << "guess" << endl;
+		cin >> cur;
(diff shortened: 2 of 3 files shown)


SI/Sem.03/Pract.03/Task04.cpp
```diff
@@ -0,0 +1,15 @@
+#include <iostream>
+using namespace std;
+int main()
+{
+	int m, n;
+	cin >> m >> n;
```
Justsvetoslavov, line 6: Check if m and n are positive.

SI/Sem.03/Pract.03/Task04.cpp
```diff
@@ -0,0 +1,15 @@
...
+	for (int i = 1; i <= m * n;) {
```
Justsvetoslavov, line 7: Revisit this solution.
It can be done in a single loop.

SI/Sem.03/Pract.03/Task06.cpp
```diff
@@ -0,0 +1,39 @@
+#include <iostream>
+#include <vector>
+using namespace std;
+int main()
+{
+	//i am using vector because I can't think of a
```
Justsvetoslavov, line 6: That is the point. No arrays are allowed. 
Revisit this task.

SI/Sem.03/Pract.03/Task05.cpp
```diff
@@ -0,0 +1,26 @@
+#include <iostream>
+using namespace std;
+int main()
+{
+	int n;
+	cout << "enter initial number" << endl;
+	cin >> n;
```
Justsvetoslavov, line 7: Check if the number is in the interval.

SI/Sem.03/Pract.03/Task05.cpp
```diff
@@ -0,0 +1,26 @@
...
+	int cur = -1;
+	int guesses = 0;
+	while (true) {
```
Justsvetoslavov, line 10: avoid while (true).
- what is the condition?
- will we always have at least 1 try?
- do we need to check for other conditions?

Justsvetoslavov (changes requested):

PR #52, opened 2022-10-19
Pract.03: Add 04-05 | Atanas Hristov

DrHristov:
Atanas Hristov
6 8MI0600242

### DIFF
--- a/SI/Sem.03/Pract.03/Task04.cpp
+++ b/SI/Sem.03/Pract.03/Task04.cpp
@@ -1,0 +1,16 @@
+#include <iostream>
+
+int main() {
+	int n = 0, m = 0, product = 0;
+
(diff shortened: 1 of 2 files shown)


SI/Sem.03/Pract.03/Task04.cpp
```diff
@@ -0,0 +1,16 @@
+#include <iostream>
+
+int main() {
+	int n = 0, m = 0, product = 0;
+
+	std::cin >> n >> m;
```
Justsvetoslavov, line 6: Check if n and m are positive numbers.
Product is not necessary.

Justsvetoslavov (approved):
Some minor changes may be done, but overall:
Kudos 👏

PR #53, opened 2022-10-19
Pract.03: Add 13-14 | Miroslav Stoyanovski

mir4oni3:
Miroslav Stoyanovski
6 5MI0600195

### DIFF
--- a/SI/Sem.03/Pract.03/Task13.cpp
+++ b/SI/Sem.03/Pract.03/Task13.cpp
@@ -1,0 +1,29 @@
+#include <iostream>
+using namespace std;
+int main()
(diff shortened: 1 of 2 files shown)


SI/Sem.03/Pract.03/Task13.cpp
```diff
@@ -0,0 +1,29 @@
+#include <iostream>
```
dejansvetoslavov, line 1: Remove '+' in the output when input is 1

dejansvetoslavov (changes requested):

GeorgiTerziev02:

PR #61, opened 2022-10-21
Pract 03: Add 04-05 | Militsa Toncheva

milicavilica:
Militsa Toncheva
Group 6 2MI0600216

### DIFF
--- a/SI/Sem.03/Pract.03/Task04.cpp
+++ b/SI/Sem.03/Pract.03/Task04.cpp
@@ -1,0 +1,26 @@
+#include <iostream>
+
+using namespace std;
+
+int main()
+{
+	int m, n;
+	cin >> m >> n;
+	if (m <= 0 || n <= 0)
+	{
+		cout << "m and n must be larger than 0";
+	}
+	else
+	{
+		int p = m * n;
+
+		for (int i = 0; i < p; i++)
+		{
+			if (i % n == 0)
+			{
+				cout << endl;
+			}
+			cout << i + 1 << ' ';
+		}
+	}
+}
--- a/SI/Sem.03/Pract.03/Task05.cpp
+++ b/SI/Sem.03/Pract.03/Task05.cpp
@@ -1,0 +1,41 @@
+#include <iostream>
+
+using namespace std;
+
+int main()
+{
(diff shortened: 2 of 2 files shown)


SI/Sem.03/Pract.03/Task04.cpp
```diff
@@ -0,0 +1,26 @@
...
+	cin >> m >> n;
+	if (m <= 0 || n <= 0)
+	{
+		cout << "m and n must be larger than 0";
+	}
+	else
+	{
+		int p = m * n;
+
+		for (int i = 0; i < p; i++)
```
Justsvetoslavov, line 17: we can start from 1 and condition will change to <=.
In addition, p is not necessary.

SI/Sem.03/Pract.03/Task04.cpp
```diff
@@ -0,0 +1,26 @@
...
+		{
+			if (i % n == 0)
+			{
+				cout << endl;
+			}
+			cout << i + 1 << ' ';
```
Justsvetoslavov, line 23: From the above changes , we will swap the if and the result and remove the +1 on the i

SI/Sem.03/Pract.03/Task05.cpp
```diff
@@ -0,0 +1,41 @@
+#include <iostream>
+
+using namespace std;
+
+int main()
+{
+	int num = 0;
+	cout << "Enter a number: ";
+	cin >> num;
+	if (num < 0 || num > 100)
```
Justsvetoslavov, line 10: Same as above.
Will return 1 and will have only the if statement.

SI/Sem.03/Pract.03/Task05.cpp
```diff
@@ -0,0 +1,41 @@
...
+	{
+		cout << "the number must be between 0 and 100";
+	}
+	else
+	{
+		int counter = 0;
+		while (true)
```
Justsvetoslavov, line 17: Re-think to remove the while (true).
Hint: use do-while.

Justsvetoslavov (changes requested):

GeorgiTerziev02:

PR #62, opened 2022-10-21
Pract.03: Add 04-06 | Vladimir Velikov

thevlad0:
Vladimir Velikov
2 0MI0600204

### DIFF
--- a/SI/Sem.03/Pract.03/Task 04 - 06/Task04.cpp
+++ b/SI/Sem.03/Pract.03/Task 04 - 06/Task04.cpp
@@ -1,0 +1,29 @@
+#include <iostream>
+
+using namespace std;
+
+int main()
+{
+    unsigned int m, n, k = 0;
+    cin >> m >> n;
+    cout << endl;
+    for (int i = 1; i <= m; i++)
+    {
+        for (int j = 1; j <= n; j++)
+        {
+            k++;
+
+            // The if and the if-else are used in order to 
+            // fix the alignment of 2-digit numbers
+            if(k <= 9)
+                cout << " " << k << " ";
+            else if (k >= 10)
+                cout << k << " ";
+
+            if(j == m*n)
+                break;
+        }
+        cout << endl;
+    }
+    
+}
--- a/SI/Sem.03/Pract.03/Task 04 - 06/Task05.cpp
+++ b/SI/Sem.03/Pract.03/Task 04 - 06/Task05.cpp
@@ -1,0 +1,34 @@
+#include <iostream>
+
+using namespace std;
+
+int main()
+{
+    int n;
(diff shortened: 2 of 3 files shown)


SI/Sem.03/Pract.03/Task 04 - 06/Task04.cpp
```diff
@@ -0,0 +1,29 @@
+#include <iostream>
+
+using namespace std;
+
+int main()
+{
+    unsigned int m, n, k = 0;
```
Justsvetoslavov, line 7: Good solutions with handling alignment of 2-digit numbers.
However, if we exceed them, we will need to adjust alignment again(3-digits etc).
Kudos tho :clap:

SI/Sem.03/Pract.03/Task 04 - 06/Task05.cpp
```diff
@@ -0,0 +1,34 @@
...
+        cout << "Invalid input" << endl;
+    else
+    {
+        int guess, guesses = 0;
+        do 
+        {
+            cout << "Take a guess: " << endl;
+            cin >> guess;
+            guesses++;
+            if(guess == n)
```
Justsvetoslavov, line 21: this is not necessary. Just move it after the loop.

SI/Sem.03/Pract.03/Task 04 - 06/Task05.cpp
```diff
@@ -0,0 +1,34 @@
+#include <iostream>
+
+using namespace std;
+
+int main()
+{
+    int n;
+    cout << "Enter a number between 0 and 100: ";
+    cin >> n;
+    cout << endl;
```
Justsvetoslavov, line 10: Not necessary in my opinion.

SI/Sem.03/Pract.03/Task 04 - 06/Task06.cpp
```diff
@@ -0,0 +1,19 @@
+#include <iostream>
+
+using namespace std;
+
+int main()
+{
+    int n, ogSum = 0, repSum = 0;
```
Justsvetoslavov, line 7: ogSum/repSum , I would suggest to review the names and pick a more suitable ones.

Justsvetoslavov (approved):
Slight changes that would be lovely to be addressed.
However, the additional folder in Pract.03 is unnecessary.
In addition, I would suggest you do not skip brackets {} for conditional statements or loop whenever it is a 1 liner(in legacy it could lead to error).
Overall, a good PR. Kudos x2 👏

Justsvetoslavov:
Closing due to additional folder.
We cannot merge or cherry-pick this way.

PR #66, opened 2022-10-22
Prakt.03: Add 04-06 | Margarita Gencheva

Margarita-Gencheva2003:


Justsvetoslavov:
Edit the description according to the PR rules.

SI/Sem.03/Pract.03/Таsк05.cpp
```diff
@@ -0,0 +1,22 @@
...
+	int number = 1, testNum = 0, count = 0;
+	std::cout << "Input number between 1 and 100: ";
+	std::cin >> number;
+	if (number >= 1 && number <= 100) {
+		do {
+			std::cin >> testNum;
+			if (testNum > number) {
+				std::cout << testNum << " > " << std::endl;
+			}
+			if (testNum < number) {
```
plamena-ilieva, line 12: If the first condition is true, you do not need to check the second one too. Use _else if_.

SI/Sem.03/Pract.03/Таsк06.cpp
```diff
@@ -0,0 +1,27 @@
+#include <iostream>
+int main() {
+	int num = 0, count = 0;
+	do {//checks if the number is possitive or 0
+		std::cin >> num;
+		if (num < 0) {
+			std::cout << "Insert positive number: " << std::endl;
+		}
+		if (num == 0) break;
```
plamena-ilieva, line 9: You do not need a break right before the end of the body - if num <= 0, the condition on line 10 will break the loop. That is why you do not need any other if-statements.

SI/Sem.03/Pract.03/Таsк06.cpp
```diff
@@ -0,0 +1,27 @@
...
+		count++;
+		if (count <= 2) {
+			std::cout << "Insert again " << num << "." << std::endl;
+		}
+		else {
+			count = 0;
+		}
+		
+	}
+}
```
plamena-ilieva, line 27: You probably misunderstood the task, so you will need to rewrite it. The user is never supposed to enter zero. You have to find the missing number in a sequence.

SI/Sem.03/Pract.03/Таsк06.cpp
```diff
@@ -0,0 +1,27 @@
+#include <iostream>
+int main() {
+	int num = 0, count = 0;
+	do {//checks if the number is possitive or 0
+		std::cin >> num;
+		if (num < 0) {
+			std::cout << "Insert positive number: " << std::endl;
```
plamena-ilieva, line 7: Use "enter" instead of "insert".

plamena-ilieva (changes requested):

GeorgiTerziev02:

PR #68, opened 2022-10-22
Pract 03: Add 07-09 | Militsa Toncheva

milicavilica:
Militsa Toncheva
Group 6 2MI0600216

### DIFF
--- a/SI/Sem.03/Pract.03/Task07.cpp
+++ b/SI/Sem.03/Pract.03/Task07.cpp
@@ -1,0 +1,26 @@
+#include <iostream>;
+
+using namespace std;
+
+int main()
+{
+	int n;
+	cin >> n;
+	if (n <= 0)
+	{
+		cout << "n must be larger than 0!";
+	}
+	else
+	{
+		int a = 3;
+		int b = 4;
+		int c = 5;		
+		while (c < n)
+		{
+			cout << a << ' ' << b << ' ' << c << endl;
+			a *= 2;
+			b *= 2;
+			c *= 2;
+		}
+	}
+}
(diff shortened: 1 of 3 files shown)


SI/Sem.03/Pract.03/Task07.cpp
```diff
@@ -0,0 +1,26 @@
...
+	if (n <= 0)
+	{
+		cout << "n must be larger than 0!";
+	}
+	else
+	{
+		int a = 3;
+		int b = 4;
+		int c = 5;		
+		while (c < n)
```
GeorgiTerziev02, line 18: Wrong solution: if you have input 13, you will skip the output 5, 12, 13.

SI/Sem.03/Pract.03/Task09.cpp
```diff
@@ -0,0 +1,36 @@
...
+int main()
+{
+	int n;
+	cout << "Вход: ";
+	cin >> n;
+	if (n > 3)
+	{
+		int num1 = 1;
+		int num2 = 1;
+		cout << "Изход: " << num1 << ' ' << num2 << ' ';
```
GeorgiTerziev02, line 14: Forget about bulgarian in your code 😃

SI/Sem.03/Pract.03/Task09.cpp
```diff
@@ -0,0 +1,36 @@
...
+		int nextNum = 0;
+		for (int i = 1; ; i++)
```
GeorgiTerziev02, line 16: Looks like a while loop will be better here, you are not using the i variable

GeorgiTerziev02 (changes requested):

GeorgiTerziev02:

PR #77, opened 2022-10-26
Pract 03: Add 10 - 12 | Militsa Toncheva

milicavilica:
Militsa Toncheva
Group 6 2MI0600216


Justsvetoslavov:
Example solutions are uploaded.
You can view them [here](https://github.com/Justsvetoslavov/Introduction_to_programming/tree/master/SI/Sem.03/Pract.03).

PR #86, opened 2022-10-26
Pract.04: Add 04-06 | Momchil Uzunov

momouzunov2003:
Momchil Uzunov
6 1MI0600146

### DIFF
--- a/SI/Sem.04/Pract.04/Task 04.cpp
+++ b/SI/Sem.04/Pract.04/Task 04.cpp
@@ -1,0 +1,37 @@
+#include <iostream>
+
+int abs(int num)
+{
+    if (num >= 0)
+    {
+        return num;
+    }
+    else
+    {
+        return -num;
+    }
+}
+
+double fabs(double num)
+{
+    if (num >= 0)
+    {
+        return num;
+    }
+    else
+    {
+        return -num;
+    }
+}
+
+int main()
+{
+    int number;
+    double number2;
+    std::cout << "Please enter a whole number: ";
+    std::cin >> number;
+    std::cout << "Please enter a decimal number: ";
+    std::cin >> number2;
+    std::cout << std::endl << "The absolute value of the whole number is: " << abs(number);
+    std::cout << std::endl << "The absolute value of the decimal number is: " << fabs(number2);
+}
--- a/SI/Sem.04/Pract.04/Task 04.cpp
+++ b/SI/Sem.04/Pract.04/Task 04.cpp
@@ -2,26 +2,14 @@
 
 int abs(int num)
 {
-    if (num >= 0)
-    {
-        return num;
-    }
-    else
-    {
-        return -num;
-    }
+    int result=(num >= 0) ? num :  -num;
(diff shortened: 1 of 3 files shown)


SI/Sem.04/Pract.04/Task 06.cpp
```diff
@@ -0,0 +1,34 @@
...
+    }
+    else
+    {
+        return false;
+    }
+}
+
+bool checkSymbol(char S)
+{
+    if ((S >= 'A' && S <= 'Z') || (S>='a' && S<='z'))
```
GeorgiTerziev02, line 17: (S >= 'A' && S <= 'Z') || (S>='a' && S<='z') is a boolean expression, so it returns a boolean value that you can directly return

SI/Sem.04/Pract.04/Task 04.cpp
```diff
@@ -0,0 +1,37 @@
+#include <iostream>
+
+int abs(int num)
+{
+    if (num >= 0)
+    {
```
GeorgiTerziev02, line 6: You can use a ternary operator here

GeorgiTerziev02 (changes requested):
Some refactoring needed

momouzunov2003:
Updated tasks 04 and 06.

SI/Sem.04/Pract.04/Task 04.cpp
```diff
@@ -0,0 +1,25 @@
+#include <iostream>
+
+int abs(int num)
+{
+    int result=(num >= 0) ? num :  -num;
+    return result;
```
GeorgiTerziev02, line 6: No need of a variable here. You can directly return

SI/Sem.04/Pract.04/Task 04.cpp
```diff
@@ -0,0 +1,25 @@
...
+}
+
+double fabs(double num)
+{
+   int result = (num >= DBL_EPSILON) ? num : -num;
+    return result;
```
GeorgiTerziev02, line 12: Not working correctly. You are assigning a double to int so you are returning the int that will be converted to double

SI/Sem.04/Pract.04/Task 06.cpp
```diff
@@ -0,0 +1,20 @@
+#include <iostream>
+
+bool isCapitalLetter(char S)
+{
+    return (S >= 'A' && S <= 'Z');
```
GeorgiTerziev02, line 5: No need of brackets

SI/Sem.04/Pract.04/Task 06.cpp
```diff
@@ -0,0 +1,20 @@
...
+}
+
+bool checkSymbol(char S)
+{
+    return ((S >= 'A' && S <= 'Z') || (S >= 'a' && S <= 'z'));
```
GeorgiTerziev02, line 10: No need of the first brackets

GeorgiTerziev02 (changes requested):

PR #87, opened 2022-10-26
Pract.04: Add 01-03 | Presiyana Borisova

PresiyanaBB:
Presiyana Borisova
6 3MI0600240

### DIFF
--- a/SI/Sem.04/Pract.04/Task02.cpp
+++ b/SI/Sem.04/Pract.04/Task02.cpp
@@ -1,0 +1,25 @@
+using namespace std;
+#include <iostream>
+
+bool isEven(int num)
+{
+    return num % 2 == 0 ? true : false;
+}
+
+int main()
+{
+    int num;
+
+    cout << "Please enter 1 number: ";
+
+    cin >> num;
+
+    if (isEven(num))
+    {
+        cout << num << " is even";
+        return 0;
+    }
+
+    cout << num <<" is odd";
+    return 0;
+}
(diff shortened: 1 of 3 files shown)


SI/Sem.04/Pract.04/Task02.cpp
```diff
@@ -0,0 +1,25 @@
+using namespace std;
+#include <iostream>
+
+bool isEven(int num)
+{
+    return num % 2 == 0 ? true : false;
```
GeorgiTerziev02, line 6: num % 2 == 0 is a boolean expression, so it returns a boolean value that you can directly return

SI/Sem.04/Pract.04/Task03.cpp
```diff
@@ -0,0 +1,36 @@
+using namespace std;
+#include <iostream>
+
+bool isPositive(int num)
+{
+    return num > 0 ? true : false;
```
GeorgiTerziev02, line 6: num > 0 is a boolean expression, so it returns a boolean value that you can directly return
return num > 0

SI/Sem.04/Pract.04/Task03.cpp
```diff
@@ -0,0 +1,36 @@
...
+}
+
+bool isNegative(int num)
+{
+    return num < 0 ? true : false;
```
GeorgiTerziev02, line 11: num < 0 is a boolean expression, so it returns a boolean value that you can directly return

GeorgiTerziev02 (changes requested):
Some refactoring needed

PR #88, opened 2022-10-26
Pract.04 task01 03

Viktoria035:
Viktoria Koleva
6   5MI0600139

### DIFF
--- a/SI/Sem.04/Pract.04/Task02.cpp
+++ b/SI/Sem.04/Pract.04/Task02.cpp
@@ -1,0 +1,21 @@
+#include <iostream>
+int isEven(int num)
+{
+	if (num % 2 == 0)
+	{
+		std::cout << "Your number is even - "<<" ";
+
+	}
+	else
+	{
+		std::cout << "Your number is odd - "<<" ";
+	}
+	return num;
+}
+int main()
+{
+	int num = 0;
+	std::cout << "Please enter a number: ";
+	std::cin >> num;
+	std::cout << isEven(num);
+}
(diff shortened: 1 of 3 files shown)


SI/Sem.04/Pract.04/Task02.cpp
```diff
@@ -0,0 +1,21 @@
+#include <iostream>
+int isEven(int num)
```
GeorgiTerziev02, line 2: Should return bool

SI/Sem.04/Pract.04/Task03.cpp
```diff
@@ -0,0 +1,35 @@
+#include <iostream>
+void isPositive(int numberP)
```
GeorgiTerziev02, line 2: Should return bool

SI/Sem.04/Pract.04/Task03.cpp
```diff
@@ -0,0 +1,35 @@
...
+{
+	if (numberP >= 0)
```
GeorgiTerziev02, line 4: zero is not positive

SI/Sem.04/Pract.04/Task03.cpp
```diff
@@ -0,0 +1,35 @@
...
+	{
+		std::cout << "Your number is positive";
+	}
+	else
+	{
+		std::cout << "Your number is negative";
+	}
+	
+}
+void isNegative(int numberN)
```
GeorgiTerziev02, line 14: should return bool

GeorgiTerziev02 (changes requested):

Justsvetoslavov:
Follow PR rules for PR title.

Closing due to merged solutions.
Address comments and contact us if you have questions.

PR #93, opened 2022-10-26
Prakt03 task10-12 Stefka Lyaskalieva

stefka04:
Stefka Lyaskalieva
4MI0600158 Group 2

### DIFF
--- a/SI/Sem.03/Pract.03/Prakt3Task10.cpp
+++ b/SI/Sem.03/Pract.03/Prakt3Task10.cpp
@@ -1,0 +1,42 @@
+#include <iostream>
+#include <cmath>
+using namespace std;
+int main()
+{
+    int num, result,  digits, firstDigit;
+    cout << "Enter a number: ";
+    cin >> num;
+    result = 0;
+    int counter = 0;
+    num != 0 ? digits = log10(num) + 1 : digits = 1;
+    
+    if (digits % 2 != 0)
+    {
+        for (int i = 1; i <= digits; i++)
+        {
+            if (i != digits / 2 + 1)
+            {
+                result = result + (num % 10) * pow(10, counter);
+                counter++;
+            }
+
+            num = num / 10;
+        }
+    }
+    else
+    {
+        for (int i = 1; i <= digits; i++)
+        {
+            if (i != digits / 2 + 1 && i != digits / 2)
+            {
+                result = result + (num % 10) * pow(10, counter);
+                counter++;
+            }
+
+            num = num / 10;
+        }
+    }
+    cout << result << ", " << result + 1;
+    return 0;
+}
+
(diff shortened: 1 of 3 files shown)


SI/Sem.03/Pract.03/Prakt3Task10.cpp
```diff
@@ -0,0 +1,42 @@
...
+            num = num / 10;
+        }
+    }
+    else
+    {
+        for (int i = 1; i <= digits; i++)
+        {
+            if (i != digits / 2 + 1 && i != digits / 2)
+            {
+                result = result + (num % 10) * pow(10, counter);
```
plamena-ilieva, line 32: You can avoid using _pow_ by doing _counter *= 10_ instead of _counter++_ (starting from value 1).

SI/Sem.03/Pract.03/Prakt3Task11.cpp
```diff
@@ -0,0 +1,33 @@
...
+    if (a > b)
+    {
+        temp = a;
+        a = b;
+        b = temp;
+    }
+    for(int i = a; i <= b; i++) 
+    {
+        isPrime = true;
+        for (int j = 2; j <= sqrt(i); j++)
```
plamena-ilieva, line 19: Never use _sqrt()_ in a loop condition because it calculates the square root at every iteration and this makes the program slower.

SI/Sem.03/Pract.03/Prakt3Task11.cpp
```diff
@@ -0,0 +1,33 @@
...
+        {
+            if (i % j == 0 && i != j)
+            {
+                isPrime = false;
+                break;
+            }
+        }
+        if(isPrime == true && i != 1)
```
plamena-ilieva, line 27: You can check whether the number is 1. Also, keep in mind that negative numbers are not prime.

plamena-ilieva (changes requested):

SI/Sem.03/Pract.03/Prakt3Task10.cpp
```diff
@@ -0,0 +1,42 @@
...
+#include <cmath>
+using namespace std;
+int main()
+{
+    int num, result,  digits, firstDigit;
+    cout << "Enter a number: ";
+    cin >> num;
+    result = 0;
+    int counter = 0;
+    num != 0 ? digits = log10(num) + 1 : digits = 1;
```
GeorgiTerziev02, line 11: Try not to use external libraries for things you can write yourself

SI/Sem.03/Pract.03/Prakt3Task11.cpp
GeorgiTerziev02, line 27: if(isPrime == true) is equal to if(isPrime) because isPrime is a boolean value

GeorgiTerziev02:
The correct solutions have already been uploaded, check them in the link below

PR #98, opened 2022-10-28
Pract.04: Add 07-09 | Kaloyan Peevski

kokopeevski:
Kaloyan Peevski
2 2MI0600170

### DIFF
--- a/SI/Sem.04/Pract.04/Task07.cpp
+++ b/SI/Sem.04/Pract.04/Task07.cpp
@@ -1,0 +1,23 @@
+﻿// Task07.cpp : This file contains the 'main' function. Program execution begins and ends there.
+//
+
+#include <iostream>
+using namespace std;
+
+int askUser(int from, int to)
+{
+	int number;
+	cout << "Enter a number: ";
+	cin >> number;
+	while (!(number >= from && number <= to))
+	{
+		cout << "Enter a new number: ";
+		cin >> number;
+	}
+	return number;
+}
+
+int main()
+{
+	cout << "The number is: " << askUser(10, 30) << endl;
+}
--- a/SI/Sem.04/Pract.04/Task08.cpp
+++ b/SI/Sem.04/Pract.04/Task08.cpp
@@ -1,0 +1,18 @@
+﻿// Task08.cpp : This file contains the 'main' function. Program execution begins and ends there.
+//
+
+#include <iostream>
+using namespace std;
+
+long long pow(int num, unsigned int N)
+{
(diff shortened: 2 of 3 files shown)


SI/Sem.04/Pract.04/Task08.cpp
```diff
@@ -0,0 +1,18 @@
...
+//
+
+#include <iostream>
+using namespace std;
+
+long long pow(int num, unsigned int N)
+{
+	for (int i = 1; i <= N; i++)
+	{
+		return num * N;
```
GeorgiTerziev02, line 11: This won't work
You will directly return

SI/Sem.04/Pract.04/Task08.cpp
```diff
@@ -0,0 +1,18 @@
...
+	}
+}
+
+int main()
+{
+	cout << pow(2,7);
```
GeorgiTerziev02, line 17: You are using the pow from cmath.h here

SI/Sem.04/Pract.04/Task09.cpp
```diff
@@ -0,0 +1,44 @@
...
+			return a - b;
+			break;
+		case '*':
+			return a * b;
+			break;
+		case '/':
+			if (b == 0)
+			{
+				cout << "The devisor can not be zero! Please input other number!";
+				cin >> b;
```
GeorgiTerziev02, line 24: User can again enter 0

SI/Sem.04/Pract.04/Task09.cpp
```diff
@@ -0,0 +1,44 @@
...
+			}
+			return a / b;
+			break;
+		case '%':
+			if (b == 0)
+			{
+				cout << "The devisor can not be zero! Please input other number!";
+				cin >> b;
```
GeorgiTerziev02, line 32: User can again enter 0

SI/Sem.04/Pract.04/Task07.cpp
```diff
@@ -0,0 +1,23 @@
...
+
+int askUser(int from, int to)
+{
+	int number;
+	cout << "Enter a number: ";
+	cin >> number;
+	while (!(number >= from && number <= to))
+	{
+		cout << "Enter a new number: ";
+		cin >> number;
```
GeorgiTerziev02, line 15: In this case do while is more optimal, you won't have duplicate code that way

GeorgiTerziev02 (changes requested):

PR #99, opened 2022-10-28
Pract.04: Add 01-03 | Miroslav Stoyanovski

mir4oni3:
Miroslav Stoyanovski 
6 5MI0600195

### DIFF
--- a/SI/Sem.04/Pract.04/Task02.cpp
+++ b/SI/Sem.04/Pract.04/Task02.cpp
@@ -1,0 +1,14 @@
+#include <iostream>
+
+bool isEven(int n) {
+	if (n % 2 == 0) {
+		return true;
+	}
+	return false;
+}
+int main() {
+	int n;
+	std::cout << "number:" << std::endl;
+	std::cin >> n;
+	std::cout << std::boolalpha << isEven(n);
+}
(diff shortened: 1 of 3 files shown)


SI/Sem.04/Pract.04/Task02.cpp
```diff
@@ -0,0 +1,14 @@
+#include <iostream>
+
+bool isEven(int n) {
+	if (n % 2 == 0) {
```
Justsvetoslavov, line 4: Single statement.
return n % 2 == 0;

SI/Sem.04/Pract.04/Task03.cpp
```diff
@@ -0,0 +1,22 @@
+#include <iostream>
+
+bool isPositive(int n) {
+	if (n > 0) {
```
Justsvetoslavov, line 4: Single statement.
return n > 0;

SI/Sem.04/Pract.04/Task03.cpp
```diff
@@ -0,0 +1,22 @@
...
+		return true;
+	}
+	return false;
+}
+bool isNegative(int n) {
+	if (n < 0) {
```
Justsvetoslavov, line 10: Single statement.
return n < 0;

Justsvetoslavov (approved):
Comparison operators return boolean (true/false).
Avoid if-else statements if you can write a single return statement.

Justsvetoslavov:
Closing due to merged solutions.
Kudos for the solutions. 👏

PR #100, opened 2022-10-28
Pract.04: Add 04-06 | Miroslav Stoyanovski

mir4oni3:
Miroslav Stoyanovski 
6 5MI0600195

### DIFF
--- a/SI/Sem.04/Pract.04/Task06.cpp
+++ b/SI/Sem.04/Pract.04/Task06.cpp
@@ -1,0 +1,23 @@
+#include <iostream>
+
+bool checkSymbol(char S) {
+	if ((S >= 65 && S <= 90) || (S >= 97 && S <= 122)) {
+		return true;
+	}
+	return false;
+}
+bool isCapitalLetter(char S) {
+	if (checkSymbol(S)) {
+		if (S >= 65 && S <= 90) {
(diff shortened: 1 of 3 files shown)


SI/Sem.04/Pract.04/Task06.cpp
```diff
@@ -0,0 +1,23 @@
+#include <iostream>
+
+bool checkSymbol(char S) {
+	if ((S >= 65 && S <= 90) || (S >= 97 && S <= 122)) {
```
Justsvetoslavov, line 4: Re-write this function.
Think how to be able with a single return.
Can we remove the integer values for the letters?

SI/Sem.04/Pract.04/Task06.cpp
```diff
@@ -0,0 +1,23 @@
...
+		return true;
+	}
+	return false;
+}
+bool isCapitalLetter(char S) {
+	if (checkSymbol(S)) {
```
Justsvetoslavov, line 10: Re-write this function.
Think how to be able with a single return.
Can we remove the integer values for the letters?

Justsvetoslavov (changes requested):
Also let us set return 0; as a standard for main function.

PR #103, opened 2022-10-28
Pract.04: Add 13-15 | Miroslav Stoyanovski

mir4oni3:
Miroslav Stoyanovski
6 5MI0600195

### DIFF
--- a/SI/Sem.04/Pract.04/Task13.cpp
+++ b/SI/Sem.04/Pract.04/Task13.cpp
@@ -1,0 +1,45 @@
+#include <iostream>
+#include <cmath>
+
+bool isPrime(int n) {
+	if (n == 1) {
+		return false;
+	}
+	for (int i = 2; i <= sqrt(n); i++) {
+		if (n % i == 0) {
+			return false;
+		}
+	}
+	return true;
+}
+
+void printAllPrimesTo(int n) {
+	for (int i = 2; i <= n; i++) {
+		if (isPrime(i)) {
+			std::cout << i << " ";
+		}
+	}
+	std::cout << std::endl;
+}
+
+void sumOfPrimeNumbers(int n) {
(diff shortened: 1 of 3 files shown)


SI/Sem.04/Pract.04/Task13.cpp
```diff
@@ -0,0 +1,45 @@
...
+	for (int i = 2; i <= sqrt(n); i++) {
+		if (n % i == 0) {
+			return false;
+		}
+	}
+	return true;
+}
+
+void printAllPrimesTo(int n) {
+	for (int i = 2; i <= n; i++) {
```
GeorgiTerziev02, line 17: You can only loop through the odd numbers

SI/Sem.04/Pract.04/Task15.cpp
```diff
@@ -0,0 +1,38 @@
...
+	return true;
+}
+
+void factorise(int n) {
+	if (isPrime(n)) {
+		std::cout << "number is prime." << std::endl;
+		return;
+	}
+	std::cout << n << " = ";
+	for (int i = 2; i <= n; i++) {
```
GeorgiTerziev02, line 21: You should loop to the square root of the number
Also you can optimize it a bit for the even numbers

SI/Sem.04/Pract.04/Task15.cpp
```diff
@@ -0,0 +1,38 @@
+#include <iostream>
+
+bool isPrime(int n) {
+	if (n == 1) {
+		return false;
+	}
+	for (int i = 2; i <= sqrt(n); i++) {
```
GeorgiTerziev02, line 7: You can loop only through the odd numbers

GeorgiTerziev02 (approved):

PR #107, opened 2022-10-29
Pract 04: 14-15 | Kristian Petrov

Kristian2109:
Kristian Petrov
6 8MI0600201

### DIFF
--- a/SI/Sem.04/Pract.04/Task14.cpp
+++ b/SI/Sem.04/Pract.04/Task14.cpp
@@ -1,0 +1,29 @@
+#include <iostream>
+using namespace std;
+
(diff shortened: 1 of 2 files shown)


SI/Sem.04/Pract.04/Task14.cpp
```diff
@@ -0,0 +1,29 @@
+#include <iostream>
+using namespace std;
+
+int sumDigits(int number) {
+	int sum = 0;
+	while (number > 0) {
+		sum = sum + (number % 10);
```
GeorgiTerziev02, line 7: No need of brackets

GeorgiTerziev02 (approved):

PR #110, opened 2022-10-29
Pract.04: Add 12-14 | Stefka Lyaskalieva

stefka04:
Stefka Lyaskalieva 
2 4MI0600158

### DIFF
--- a/SI/Sem.04/Pract.04/Task12.cpp
+++ b/SI/Sem.04/Pract.04/Task12.cpp
@@ -1,0 +1,41 @@
+#include <iostream>
+using namespace std;
+int checkNum(int num)
+{
+	while (num < 100 || num > 100000)
+	{
+		cout << "Invalid input! Enter a new number:";
+		cin >> num;
+	}
+	return num;
+}
+int abs(int num)
+{
+	return num > 0 ? num : -num;
+}
+int closestPowOfTwo(int num)
+{   
+	int closest = 100000;
+	for(int i = 128; ; i *= 2)
+	{
+		if (abs(num - i) < closest)
+		{
+			closest = abs(num - i);
+		}
+		else
+		{
+			closest = i / 2;
+			break;
+		}
+	}
+	return closest;
+}
+int main()
+{   
+	cout << "Enter a number from 100 to 100000: ";
+	int num;
+	cin >> num;
+	cout << closestPowOfTwo(checkNum(num));
+	return 0;
+}
+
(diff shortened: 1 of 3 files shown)


SI/Sem.04/Pract.04/Task12.cpp
```diff
@@ -0,0 +1,41 @@
...
+int abs(int num)
+{
+	return num > 0 ? num : -num;
+}
+int closestPowOfTwo(int num)
+{   
+	int closest = 100000;
+	for(int i = 128; ; i *= 2)
+	{
+		if (abs(num - i) < closest)
```
GeorgiTerziev02, line 21: The if else can be simplified to be only if

SI/Sem.04/Pract.04/Task13.cpp
```diff
@@ -0,0 +1,56 @@
+#include <iostream>
+using namespace std;
+bool isPrime(int num)
+{
+	int squreRoot = sqrt(num);
```
GeorgiTerziev02, line 5: You define this variable way too early, you will use it on line 11.

SI/Sem.04/Pract.04/Task13.cpp
```diff
@@ -0,0 +1,56 @@
...
+		{
+			isPrime = false;
+			break;
+		}
+	}
+	return isPrime;
+}
+void allPrimeNums(int num)
+{
+	for (int i = 1; i < num; i++)
```
GeorgiTerziev02, line 23: Can be optimized to skip all even numbers except 2

SI/Sem.04/Pract.04/Task13.cpp
```diff
@@ -0,0 +1,56 @@
...
+			cout<< i <<" ";
+		}
+	}
+}
+
+bool isSumOfTwoPrimeNums(int num)
+{
+	bool isSum = false;
+	for (int i = 2; i < num; i++)
+	{
```
GeorgiTerziev02, line 36: Can be optimized => on line 36 if(!isPrime(i)) continue. Also you can skip all the even numbers except for two.

SI/Sem.04/Pract.04/Task14.cpp
```diff
@@ -0,0 +1,36 @@
...
+using namespace std;
+int sumOfDigits(long long num)
+{
+	int digits = 0;
+	int numCopy = num;
+	int sum = 0;
+	while (numCopy != 0)
+	{
+		numCopy /= 10;
+		digits++;
```
GeorgiTerziev02, line 11: You can directly sum the digits while you divide the number, in that way you will skip the next for loop

GeorgiTerziev02 (approved):
Some optimizations can be made on your solutions

Justsvetoslavov (approved):

PR #111, opened 2022-10-29
Pract.04: Add 15-15 | Stefka Lyaskalieva

stefka04:
Stefka Lyaskalieva
2 4MI0600158

### DIFF
--- a/SI/Sem.04/Pract.04/Task15.cpp
+++ b/SI/Sem.04/Pract.04/Task15.cpp
@@ -1,0 +1,49 @@
+#include <iostream>
+using namespace std;
+bool isPrime(int num)
+{
+	int squreRoot = sqrt(num);
+	if (num <= 1)
+	{
+		return false;
+	}
+	bool isPrime = true;
+	for (int i = 2; i <= squreRoot; i++)
+	{
+		if (num % i == 0)
+		{
+			isPrime = false;
+			break;
+		}
+	}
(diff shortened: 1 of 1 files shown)


SI/Sem.04/Pract.04/Task15.cpp
```diff
@@ -0,0 +1,49 @@
...
+	return isPrime;
+}
+void factorization(int num)
+{
+	if (isPrime(num))
+	{
+		cout << "The number is prime and can't be factorise!";
+		return;
+	}
+	for (int i = 2; i <= num; i++)
```
GeorgiTerziev02, line 28: You should loop to the square root of the number
Also you can optimize it a bit for the even numbers

SI/Sem.04/Pract.04/Task15.cpp
```diff
@@ -0,0 +1,49 @@
...
+using namespace std;
+bool isPrime(int num)
+{
+	int squreRoot = sqrt(num);
+	if (num <= 1)
+	{
+		return false;
+	}
+	bool isPrime = true;
+	for (int i = 2; i <= squreRoot; i++)
```
GeorgiTerziev02, line 11: Same as your last PR

GeorgiTerziev02 (approved):
Some optimizations can be made

Justsvetoslavov (approved):

PR #122, opened 2022-11-01
Pract.04: Add 07-09 | Georgi Stoychev

georgiStoychev17:
Georgi Stoychev 
2 7MI0600157

### DIFF
--- a/Tasks4/Tasks4/Task9.cpp
+++ b/Tasks4/Tasks4/Task9.cpp
@@ -1,0 +1,16 @@
+#include <iostream>
+
+int calculate(int a, int b, char action = '+') {
+
+	int result;
+
+	switch (action) {
+
+	case'+': result = a + b; break;
+	case '-': result = a - b; break;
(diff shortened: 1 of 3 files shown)


Tasks4/Tasks4/Task9.cpp
```diff
@@ -0,0 +1,16 @@
...
+int calculate(int a, int b, char action = '+') {
+
+	int result;
+
+	switch (action) {
+
+	case'+': result = a + b; break;
+	case '-': result = a - b; break;
+	case '*': result = a * b; break;
+	case '/': result = a / b; break;
```
plamena-ilieva, line 12: Be careful when dividing by 0.

plamena-ilieva (changes requested):

Justsvetoslavov:
Keep track of folders as other PR.

We need to merge to SI/SemX/PractX (where X is the number of the specific week).

PR #123, opened 2022-11-01
Pract.04: Add 11-13 | Georgi Stoychev

georgiStoychev17:
Georgi Stoychev 
2 7MI0600157

### DIFF
--- a/Tasks4/Tasks4/Task13.cpp
+++ b/Tasks4/Tasks4/Task13.cpp
@@ -1,0 +1,42 @@
+#include <iostream>
+
+bool isPrime(int num) {
+
+	int sqrtNum = sqrt(num);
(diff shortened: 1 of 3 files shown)


Tasks4/Tasks4/Task13.cpp
```diff
@@ -0,0 +1,42 @@
...
+
+	int sqrtNum = sqrt(num);
+	for (int i = 2; i <= sqrtNum; i++) {
+
+		if (num % i == 0) {
+			return false;
+		}
+	}
+
+	return true;
```
plamena-ilieva, line 13: If the number is 0 or 1, the function returns true.

plamena-ilieva (changes requested):

Justsvetoslavov:
PR is in wrong folders.

Keep track of that.

PR #125, opened 2022-11-01
Pract.03: Add 04-06 | Polina Doncheva

polinaDoncheva:
Polina Doncheva
6 9MI0600223

### DIFF
--- a/Task04.cpp
+++ b/Task04.cpp
@@ -1,0 +1,34 @@
+
+
+#include <iostream>
+using namespace std;
+
+int main()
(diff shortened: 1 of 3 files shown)


Justsvetoslavov (changes requested):
Removed comment.

Justsvetoslavov:
You need to commit to the specific folder.
In this case SI/Sem.03/Pract.03.

We discussed in person the PR (I believe). Closing with duplicate as solutions are uploaded.

PR #126, opened 2022-11-01
Pract.04: Add 05 - 07 | Daniel Jordanov

danielgo26:
Daniel Jordanov
6 3MI0600149

### DIFF
--- a/SI/Sem.04/Pract.04/Task07.cpp
+++ b/SI/Sem.04/Pract.04/Task07.cpp
@@ -1,0 +1,49 @@
+// Task07.cpp : This file contains the 'main' function. Program execution begins and ends there.
+//
+
+#include <iostream>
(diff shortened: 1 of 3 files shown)


SI/Sem.04/Pract.04/Task07.cpp
```diff
@@ -0,0 +1,49 @@
...
+}
+int AskUser(int from, int to)
+{
+	
+	int number = 0;
+	while (!CheckIsInInterval(number, from, to))
+	{
+		std::cout << "Enter number: ";
+		std::cin >> number;
+		if (!CheckIsInInterval(number, from, to))
```
plamena-ilieva, line 29: You check the same thing twice.

plamena-ilieva (changes requested):

PR #130, opened 2022-11-01
Pract.03: Add 07-09 | Polina Doncheva

polinaDoncheva:
Polina Doncheva
6 9MI0600223


Justsvetoslavov:
We discussed in person the PR (I believe). Closing with duplicate as solutions are uploaded.

PR #133, opened 2022-11-03
Doc Fix Pract 05 Array lesson

mir4oni3:
-fixed array where element should be 44
-added missing ;


Justsvetoslavov (approved):
@mir4oni3 Could you add the changes for IS/Sem.05 as well ?

Thank you!

GeorgiTerziev02:
What is the problem with the 44 in the array? Am I missing something?

SI/Sem.05/README.md
```diff
@@ -36,7 +36,7 @@ int arr[SIZE]; //ok!
 Достъпът става посредством индекси. Индексацията започва от 0. <br />
 
 ```c++
-int arr[] = {1, 2, 400, 4, 5}
+int arr[] = {1, 2, 400, 44, 5}
```
GeorgiTerziev02, line 39: Why?

mir4oni3:
It was 4 before I fixed it, now it is 44.

GeorgiTerziev02 (changes requested):
No need to change the 44, changing the value is on the next line. The point is that initialy on index 3 we have 4, the we change it to 44

mir4oni3:
You are right! My bad.

PR #134, opened 2022-11-03
Pract.04: Add 10-12 | Silvana Nikolova

SilvanaNikolova:
Silvana Nikolova
2 0MI0600152

### DIFF
--- a/SI/Sem.04/Pract.04/Task10.cpp
+++ b/SI/Sem.04/Pract.04/Task10.cpp
@@ -1,0 +1,39 @@
+﻿// Task10.cpp : This file contains the 'main' function. Program execution begins and ends there.
+//
+
+#include <iostream>
+using namespace std;
+
+void decToBinary(int number)
+{
+	int binaryNum[32];
+	int i = 0;
+
+	while (number > 0)
+	{
+		binaryNum[i] = number % 2;
+		number /= 2;
+		i++;
+	}
+
+	for (int j = i - 1; j >= 0; j--)
+	{
+		cout << binaryNum[j];
+	}
+}
+
+int main()
+{
+	decToBinary(99);
+}
+
+// Run program: Ctrl + F5 or Debug > Start Without Debugging menu
+// Debug program: F5 or Debug > Start Debugging menu
+
+// Tips for Getting Started: 
+//   1. Use the Solution Explorer window to add/manage files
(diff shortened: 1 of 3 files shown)


SI/Sem.04/Pract.04/Task10.cpp
```diff
@@ -0,0 +1,39 @@
+﻿// Task10.cpp : This file contains the 'main' function. Program execution begins and ends there.
+//
+
+#include <iostream>
+using namespace std;
+
+void decToBinary(int number)
+{
+	int binaryNum[32];
```
GeorgiTerziev02, line 9: Arrays are not allowed for Pract04

SI/Sem.04/Pract.04/Task10.cpp
```diff
@@ -0,0 +1,39 @@
...
+		cout << binaryNum[j];
+	}
+}
+
+int main()
+{
+	decToBinary(99);
+}
+
+// Run program: Ctrl + F5 or Debug > Start Without Debugging menu
```
GeorgiTerziev02, line 30: Don't upload these comments

SI/Sem.04/Pract.04/Task12.cpp
```diff
@@ -1,35 +1,54 @@
...
-	do {
-		cout << "Enter input: " << endl;
-		cin >> input;
-	} while (input < 100 || input > 1000000);
+int smallestNumber(int a)
+{
+    for (int i = a; i >= 0; --i)
+    {
+
+        if ((int)(log(i) / log(2)) == log(i) / log(2))
```
GeorgiTerziev02, line 13: The task should be done without using log. Also, you can optimize your solution.

SI/Sem.04/Pract.04/Task11.cpp
```diff
@@ -1,30 +1,87 @@
...
 
-	return result;
+	int finalDiv;
+	for (int i = secondDiv; i >= firstDiv; --i)
+	{
+		if (a % i == 0 && b % i == 0 && c % i == 0 && d % i == 0)
+		{
+			return i;
+		}
+	}
```
GeorgiTerziev02, line 71: Easier way is just to find the gcd of a and b, then find the gcd of c and d, finally find the gcd of the two results

GeorgiTerziev02 (changes requested):

PR #135, opened 2022-11-04
Pract.05: Add 02-04 | Kristian Petrov

Kristian2109:
Kristian Petrov
6 8MI0600201


Justsvetoslavov (changes requested):
Revise the tasks.

You need to use functions.

After implementing them with functions tag me for review

PR #136, opened 2022-11-04
Pract.05: Add 01-03 | Vladimir Velikov

thevlad0:
Vladimir Velikov
2 0MI0600204

### DIFF
--- a/SI/Sem.05/Pract.05/Task01.cpp
+++ b/SI/Sem.05/Pract.05/Task01.cpp
@@ -1,0 +1,44 @@
+#include <iostream>
+
+using namespace std;
+
+void RepetitiveNumbers(int* numbers, int size)
+{
+    for (int i = 0; i < size; i++)
+    {
+        for (int j = i + 1; j < size; j++)
+        {
+            if(numbers[i] == numbers[j])
+            {
+                cout << "First repeated number was: " << numbers[i] << endl;
+                return;
+            }
+        }
+        
+    }
+    cout << "No number was repeated in this sequence" << endl;;
+}
+
+void fill(int* array, int size)
+{
+    for (int i = 0; i < size; i++)
+    {
+        int currentNum;
+        cin >> currentNum;
+        array[i] = currentNum;
+    }
+}
+
+int main()
+{
+    int numbers[100];
+    int N;
+    cout << "How many numbers will you enter: ";
+    cin >> N;
+
+    cout << endl;
+    fill(numbers, N);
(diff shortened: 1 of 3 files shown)


SI/Sem.05/Pract.05/Task01.cpp
```diff
@@ -0,0 +1,44 @@
...
+        
+    }
+    cout << "No number was repeated in this sequence" << endl;;
+}
+
+void fill(int* array, int size)
+{
+    for (int i = 0; i < size; i++)
+    {
+        int currentNum;
```
Justsvetoslavov, line 26: cin >> array[i]

SI/Sem.05/Pract.05/Task01.cpp
```diff
@@ -0,0 +1,44 @@
...
+                cout << "First repeated number was: " << numbers[i] << endl;
+                return;
+            }
+        }
+        
+    }
+    cout << "No number was repeated in this sequence" << endl;;
+}
+
+void fill(int* array, int size)
```
Justsvetoslavov, line 22: fillArray or something like that.

Be more precise in the namings.

SI/Sem.05/Pract.05/Task01.cpp
```diff
@@ -0,0 +1,44 @@
+#include <iostream>
+
+using namespace std;
+
+void RepetitiveNumbers(int* numbers, int size)
+{
+    for (int i = 0; i < size; i++)
```
Justsvetoslavov, line 7: i < size - 1

SI/Sem.05/Pract.05/Task03.cpp
```diff
@@ -0,0 +1,37 @@
...
+    cout << endl;
+}
+
+int main()
+{
+    int N, size = 0;
+    cin >> N;
+    
+    int binary[100];
+    while(N != 0)
```
Justsvetoslavov, line 27: Revise this task.
Write a toBinary function ,which returns the number in binary.
Think of a solutions without swapping

SI/Sem.05/Pract.05/Task02.cpp
```diff
@@ -0,0 +1,80 @@
+#include <iostream>
```
Justsvetoslavov, line 1: Good initial solution.

Can we think how to do it without x2 more arrays ?

Justsvetoslavov (changes requested):

PR #138, opened 2022-11-04
Pract.04: Add 09-10 | Zeyneb Ahmed

zeynebAhmed:
Zeyneb Ahmed
2 2MI0600257

### DIFF
--- a/SI/Sem.04/Pract.04/Task09.cpp
+++ b/SI/Sem.04/Pract.04/Task09.cpp
@@ -1,0 +1,35 @@
+
+
+#include <iostream>
+using namespace std;
+int calculate(int a, int b, char action) {
+
+	int result;
+
+	switch (action)	{
+	case '-' :
+		result = a - b;
+			break;
+	case '*':
+		result = a * b;
+		break;
+	case '/':
+		result = a / b;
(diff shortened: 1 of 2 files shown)


SI/Sem.04/Pract.04/Task09.cpp
```diff
@@ -0,0 +1,35 @@
...
+			break;
+	case '*':
+		result = a * b;
+		break;
+	case '/':
+		result = a / b;
+		break;
+	case '%':
+		result = a % b;
+		break;
```
GeorgiTerziev02, line 21: check if b is zero

SI/Sem.04/Pract.04/Task10.cpp
```diff
@@ -0,0 +1,40 @@
...
+    if (N == 0) {
+        result = 1;
+    }
+    else if (N > 0) {
+        for (N; N >= 1; N--) {
+            result *= num;
+        }
+    }
+    return result;
+}
```
GeorgiTerziev02, line 18: Same as the other PR

GeorgiTerziev02 (changes requested):

PR #139, opened 2022-11-04
Pract.04: Add 06-08 | Zeyneb Ahmed

zeynebAhmed:
Zeyneb Ahmed
 2 2MI0600257

### DIFF
--- a/SI/Sem.04/Pract.04/Task06.cpp
+++ b/SI/Sem.04/Pract.04/Task06.cpp
@@ -1,0 +1,30 @@
+
+#include <iostream>
+using namespace std;
+
+bool checkSymbol(char S)
+{
+	return ((S >= 65 && S <= 90) || (S >= 97 && S <= 122));
+}
+
+
+bool isCapitalLetter(char S)
+{
+	
+
+	if (checkSymbol)
+	{
+		return (S >= 65 && S <= 90);
+	}
+	else
+	{
+		return 0;
+	}
+		
+}
+
+int main()
+{
+	
+	cout << isCapitalLetter('A');
+}
--- a/Task07.cpp
+++ b/Task07.cpp
@@ -1,0 +1,26 @@
+#include <iostream>
+using namespace std;
+
+int askUser(int from, int to) {
+    int num;
+    cin >> num;
+
+    if (num > from && num < to) {
+        return num;
+    }
+    else {
+        while (num <= from || num >= to) {
+            cout << "?\n";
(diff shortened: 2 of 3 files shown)


SI/Sem.04/Pract.04/Task06.cpp
```diff
@@ -0,0 +1,30 @@
+
+#include <iostream>
+using namespace std;
+
+bool checkSymbol(char S)
+{
+	return ((S >= 65 && S <= 90) || (S >= 97 && S <= 122));
```
GeorgiTerziev02, line 7: No need of the outer brackets

SI/Sem.04/Pract.04/Task06.cpp
```diff
@@ -0,0 +1,30 @@
...
+	
+
+	if (checkSymbol)
+	{
+		return (S >= 65 && S <= 90);
+	}
+	else
+	{
+		return 0;
+	}
```
GeorgiTerziev02, line 22: No need of the else block, you can directly return 0

SI/Sem.04/Pract.04/Task06.cpp
```diff
@@ -0,0 +1,30 @@
...
+}
+
+
+bool isCapitalLetter(char S)
+{
+	
+
+	if (checkSymbol)
+	{
+		return (S >= 65 && S <= 90);
```
GeorgiTerziev02, line 17: No need of brackets

Task07.cpp
```diff
@@ -0,0 +1,26 @@
...
+
+    if (num > from && num < to) {
+        return num;
+    }
+    else {
+        while (num <= from || num >= to) {
+            cout << "?\n";
+            cin >> num;
+        }
+    }
```
GeorgiTerziev02, line 16: You can do it cleaner with a do-while loop

Task08.cpp
```diff
@@ -0,0 +1,25 @@
...
+
+    if (N == 0) {
+        result = 1;
+    }
+    else if (N > 0) {
+        for (N; N >= 1; N--) {
+            result *= num;
+        }
+    }
+    return result;
```
GeorgiTerziev02, line 16: A more clear approach of writing this code
if(n == 0) return 1
if(N < 0) return 0
// then you do the for loop with the result

GeorgiTerziev02 (changes requested):
Looks good. Some code can be refactored.

PR #141, opened 2022-11-05
Pract.05: Add 01-03 | Daria Lazarova

DariaLaz:
Daria Lazarova
2 8MI0600140

### DIFF
--- a/SI/Sem.05/task01.cpp
+++ b/SI/Sem.05/task01.cpp
@@ -1,0 +1,18 @@
+﻿#include <iostream>
+
+int main()
+{
+	int arr[] = {0, 8, 7, 6, 4, 7, 9, 7, 8, 6};
+
+	for (int i = 0; i < 10; i++)
+	{
+		for (int j = 0; j < 10; j++)
+		{
+			if (arr[i] == arr[j] && i != j)
+			{
+				std::cout << arr[i] << std::endl;
+				return 0;
+			}
+		}
+	}
+}
--- a/SI/Sem.05/task02.cpp
+++ b/SI/Sem.05/task02.cpp
@@ -1,0 +1,80 @@
+﻿#include <iostream>
+
+void PrintArray(int arr[], int size) {
+	for (int i = 0; i < size; i++)
+	{
+		std::cout << arr[i] << " ";
+	}
+}
+
+void SordDescending(int arr[], int size) 
+{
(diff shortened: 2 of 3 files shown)


SI/Sem.05/task01.cpp
```diff
@@ -0,0 +1,18 @@
...
+
+int main()
+{
+	int arr[] = {0, 8, 7, 6, 4, 7, 9, 7, 8, 6};
+
+	for (int i = 0; i < 10; i++)
+	{
+		for (int j = 0; j < 10; j++)
+		{
+			if (arr[i] == arr[j] && i != j)
```
GeorgiTerziev02, line 11: Start j from i + 1 so that you can skip the check for i != j

SI/Sem.05/task02.cpp
```diff
@@ -0,0 +1,80 @@
...
+{
+	for (int i = 0; i < (size - 1); ++i)
+	{
+		for (int j = 0; j < size - 1 - i; ++j)
+		{
+			if (arr[j] > arr[j + 1])
+			{
+				int temp = arr[j + 1];
+				arr[j + 1] = arr[j];
+				arr[j] = temp;
```
GeorgiTerziev02, line 20: Swap func

SI/Sem.05/task02.cpp
```diff
@@ -0,0 +1,80 @@
...
+{
+	for (int i = 0; i < (size - 1); ++i)
+	{
+		for (int j = 0; j < size - 1 - i; ++j)
+		{
+			if (arr[j] < arr[j + 1])
+			{
+				int temp = arr[j + 1];
+				arr[j + 1] = arr[j];
+				arr[j] = temp;
```
GeorgiTerziev02, line 35: Swap func

SI/Sem.05/task02.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+
+int main()
+{
+	int arr[] = { 0, 8, 7, 6, 4, 3, 9, 7, 8, 6 };
+
+	int sortedArr[10];
+
+	int oddArr[10];
+	int evenArr[10];
```
GeorgiTerziev02, line 49: You solution works, but the point of the task is to use only one array

GeorgiTerziev02 (changes requested):

PR #142, opened 2022-11-05
Pract.05: Add 07-08 | Daria Lazarova

DariaLaz:
Daria Lazarova
2 8MI0600140

### DIFF
--- a/SI/Sem.05/task07.cpp
+++ b/SI/Sem.05/task07.cpp
@@ -1,0 +1,32 @@
+﻿#include <iostream>
+
+void PrintMatrix(int arr[][3], int rows) {
+    for (int row = 0; row < rows; row++)
+    {
+        for (int col = 0; col < 3; col++)
+        {
+            std::cout << arr[row][col] << " ";
+        }
+        std::cout << std::endl;
+    }
+}
+
+void transp(int arr[][3], int newArr[][3], int rows) {
+
+    for (int row = 0; row < rows; row++)
+    {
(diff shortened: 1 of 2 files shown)


SI/Sem.05/task07.cpp
```diff
@@ -0,0 +1,32 @@
...
+void transp(int arr[][3], int newArr[][3], int rows) {
+
+    for (int row = 0; row < rows; row++)
+    {
+        for (int col = 0; col < 3; col++)
+        {
+            newArr[col][row] = arr[row][col];
+        }
+    }
+}
```
GeorgiTerziev02, line 23: The functions should work with matrixes NxN, currently it is working with Nx3. Solution: parameters of the functions should be int**

GeorgiTerziev02, line 23: This solutions works, but the point of the task is to do it with only one array

SI/Sem.05/task08.cpp
```diff
@@ -0,0 +1,40 @@
+﻿#include <iostream>
+
+void Multiply(int arr1[][2], int arr2[][4], int multiplArr[3][4]) {
```
GeorgiTerziev02, line 3: Same as last task pass the arrays as int**

GeorgiTerziev02, line 3: You should if the multiplication of the two matrixes is possible

GeorgiTerziev02 (changes requested):

PR #143, opened 2022-11-05
Pract.05: Add 04-06 | Daria Lazarova

DariaLaz:
Daria Lazarova
2 8MI0600140

### DIFF
--- a/SI/Sem.05/task04.cpp
+++ b/SI/Sem.05/task04.cpp
@@ -1,0 +1,32 @@
+﻿#include <iostream>
+
+//int getLengthOfANum(int n) {
+//    int counter = 0;
+//    while (n != 0)
+//    {
+//        n / 10;
+//        counter++;
+//    }
+//    return counter;
+//}
+
+int main()
+{
+    /*int n;
+    std::cout << "N = ";
+    std::cin >> n;*/
+    
+    int num[] = { 1, 2, 4, 8, 4, 7, 1 };
+    int size = 7;
+    for (int i = 0; i < size / 2; i++)
+    {
+        if (num[i] != num[size - i - 1])
+        {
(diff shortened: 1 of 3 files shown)


SI/Sem.05/task04.cpp
```diff
@@ -0,0 +1,32 @@
...
+//    int counter = 0;
+//    while (n != 0)
+//    {
+//        n / 10;
+//        counter++;
+//    }
+//    return counter;
+//}
+
+int main()
```
GeorgiTerziev02, line 13: Create a function that accepts an array and length and returns boolean

SI/Sem.05/task04.cpp
```diff
@@ -0,0 +1,32 @@
...
+
+//int getLengthOfANum(int n) {
+//    int counter = 0;
+//    while (n != 0)
+//    {
+//        n / 10;
+//        counter++;
+//    }
+//    return counter;
+//}
```
GeorgiTerziev02, line 11: Remove unnecessary comments

SI/Sem.05/task06.cpp
```diff
@@ -0,0 +1,33 @@
...
+    {
+        if (i % 2 == 0 && arr[i] < arr[i - 1])
+        {
+            swap(arr[i], arr[i - 1]);
+        }
+        else if (i % 2 == 1 && arr[i] > arr[i - 1])
+        {
+            swap(arr[i], arr[i - 1]);
+        }
+    }
```
GeorgiTerziev02, line 30: Create function

GeorgiTerziev02 (changes requested):
Everything looks good but the purpose of the tasks is to create functions for the main functionalities

PR #148, opened 2022-11-07
Pract.05: Add 01-03 | Stefka Lyaskalieva

stefka04:
Stefka Lyaskalieva
2 4MI0600158

### DIFF
--- a/SI/Sem.05/Pract.05/Task02.cpp
+++ b/SI/Sem.05/Pract.05/Task02.cpp
@@ -1,0 +1,56 @@
+#include <iostream>
+using namespace std;
+void swap(int& a, int& b)
+{
+	int temp = a;
+	a = b;
+	b = temp;
+}
+void SortEvenNumsAscending(int* arr, int Size)
+{
+	for (int i = 0; i < Size; i++)
+	{
+		for (int j = i + 1; j < Size; j++)
+		{
+			if (arr[i] % 2 == 0 && arr[i] > arr[j])
+			{
+				swap(arr[i], arr[j]);
+			}
+		}
+	}
+}
+void SortOddNumsDescending(int* arr, int Size)
+{
+	for (int i = 0; i < Size; i++)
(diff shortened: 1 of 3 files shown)


SI/Sem.05/Pract.05/Task02.cpp
```diff
@@ -0,0 +1,56 @@
+#include <iostream>
+using namespace std;
+void swap(int& a, int& b)
+{
+	int temp = a;
+	a = b;
+	b = temp;
+}
+void SortEvenNumsAscending(int* arr, int Size)
```
GeorgiTerziev02, line 9: Be consistent in naming you functions. (Currently swap is camelCase, and the other is PascalCase)

SI/Sem.05/Pract.05/Task03.cpp
```diff
@@ -0,0 +1,21 @@
...
+    while (num > 0)
+    {
+        binaryNum[size] = num % 2;
+        num /= 2;
+        size ++;
+    }
+    for (int i = size - 1; i >= 0; i--)
+    {
+        cout << binaryNum[i];
+    }
```
GeorgiTerziev02, line 19: Create a function

SI/Sem.05/Pract.05/Task02.cpp
```diff
@@ -0,0 +1,56 @@
...
+	{
+		for (int j = i + 1; j < Size; j++)
+		{
+			if (arr[i] % 2 != 0 && arr[i] < arr[j])
+			{
+				swap(arr[i], arr[j]);
+			}
+		}
+	}
+}
```
GeorgiTerziev02, line 34: This solutions works. Maybe it will be more optimal to first put the even numbers to the left and the odd to the right and then sort every half? @Justsvetoslavov

GeorgiTerziev02 (changes requested):

PR #149, opened 2022-11-07
Pract.05: Add 04-06 | Stefka Lyaskalieva

stefka04:
Stefka Lyaskalieva
2 4MI0600158

### DIFF
--- a/SI/Sem.05/Pract.05/Task04.cpp
+++ b/SI/Sem.05/Pract.05/Task04.cpp
@@ -1,0 +1,28 @@
+#include <iostream>
+using namespace std;
+bool isPalindrome(int num)
+{
+    const int Size = 5; 
+    int arr[Size];
+    for (int i = 0; i < Size; i++)
+    {
+        arr[i] = num % 10; 
+        num /= 10;
+    }
+    for (int i = 0; i < Size/2; i++)
+    {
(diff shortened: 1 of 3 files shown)


SI/Sem.05/Pract.05/Task04.cpp
```diff
@@ -0,0 +1,28 @@
...
+    }
+    for (int i = 0; i < Size/2; i++)
+    {
+        if (arr[i] != arr[Size - i - 1])
+        {
+            return false;
+        }
+    }
+    return true;
+}
```
GeorgiTerziev02, line 20: This should be function that accepts int[] and length

SI/Sem.05/Pract.05/Task05.cpp
```diff
@@ -0,0 +1,53 @@
...
+			else if (arr1[i] < arr2[j])
+			{
+				i++;
+			}
+			else if (arr1[i] > arr2[j])
+			{
+				j++;
+			}
+	} 
+}
```
GeorgiTerziev02, line 44: Great

GeorgiTerziev02 (changes requested):
Good job. Change the function on task04 as in the comment I left.

Justsvetoslavov:
Cherry-picked 5 & 6

PR #150, opened 2022-11-07
Pract.05: Add 04-06 | Vladimir Velikov

thevlad0:
Vladimir Velikov
2 0MI0600204

### DIFF
--- a/SI/Sem.05/Pract.05/Task06.cpp
+++ b/SI/Sem.05/Pract.05/Task06.cpp
@@ -1,0 +1,58 @@
+#include <iostream>
+
+using namespace std;
+
+void fillArray(int* array, int size)
+{
+    for (int i = 0; i < size; i++)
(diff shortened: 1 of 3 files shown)


SI/Sem.05/Pract.05/Task06.cpp
```diff
@@ -0,0 +1,58 @@
...
+    }
+    cout << endl;
+}
+
+void taskOrder(int* array, int size)
+{
+    reverseBubbleSort(array, size);
+    for (int i = 1; i < size - 1; i+=2)
+        swap(array[i], array[i+1]);    
+}
```
GeorgiTerziev02, line 44: The solution works, but it can be done much easier

GeorgiTerziev02 (changes requested):

Justsvetoslavov:
Cherry picked 4th.

PR #152, opened 2022-11-08
Pract.05: Add 04-06 | Miroslav Stoyanovski

mir4oni3:
Miroslav Stoyanovski
6 5MI0600195

### DIFF
--- a/SI/Sem.05/Pract.05/Task05.cpp
+++ b/SI/Sem.05/Pract.05/Task05.cpp
@@ -1,0 +1,63 @@
+#include <iostream>
+
+int main() {
+	const int MAX_LENGTH = 100;
+	int arr[MAX_LENGTH];
+	int arr1[MAX_LENGTH];
+
+	std::cout << "enter size of first array (<=" << MAX_LENGTH << "):" << std::endl;
+	int n;
+	std::cin >> n;
+	if (n > MAX_LENGTH) {
+		std::cout << "size must be <=" << MAX_LENGTH << std::endl;
+		return 0;
+	}
+	std::cout << "enter elements(sorted):" << std::endl;
+	for (int i = 0; i < n; i++) {
+		int element;
+		std::cin >> element;
+		arr[i] = element;
+	}
+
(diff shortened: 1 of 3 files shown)


SI/Sem.05/Pract.05/Task05.cpp
```diff
@@ -0,0 +1,63 @@
...
+		}
+		//if i is in second array and if check == true, prints element
+		for (int j = 0; j < m; j++) {
+			if (arr1[j] == i && check) {
+				std::cout << i << " ";
+				break;
+			}
+		}
+		check = false;
+	}
```
GeorgiTerziev02, line 62: This is not the optimal solution for the task with sorted arrays. (Hint) One while loop with two indexes i and j that iterate through the two arrays and compare the numbers.

SI/Sem.05/Pract.05/Task06.cpp
```diff
@@ -0,0 +1,43 @@
...
+	}
+
+	//sort
+	int i, j;
+	for (i = 0; i < n - 1; i++) { 
+		for (j = 0; j < n - i - 1; j++) { 
+			if (arr[j] > arr[j + 1]) {
+				int cur = arr[j];
+				arr[j] = arr[j + 1];
+				arr[j + 1] = cur;
```
GeorgiTerziev02, line 27: swap function

SI/Sem.05/Pract.05/Task06.cpp
```diff
@@ -0,0 +1,43 @@
...
+			}
+		}
+	}
+
+	//sort by rule
+	for (int i = 1; i < n; i+=2) {
+		int cur = arr[i];
+		arr[i] = arr[i - 1];
+		arr[i - 1] = cur;
+	}
```
GeorgiTerziev02, line 37: The solution can be done without sorting, but just with one loop that iterates through the array and if two numbers(one after another) are not in the correct order - swap

GeorgiTerziev02 (changes requested):

PR #156, opened 2022-11-09
Pract.05: Add 04 - 06 | Daniel Jordanov

danielgo26:
Daniel Jordanov
6 3MI0600149

### DIFF
--- a/SI/Sem.05/Pract.05/Task04.cpp
+++ b/SI/Sem.05/Pract.05/Task04.cpp
@@ -1,0 +1,52 @@
+#include <iostream>
+bool IsPalindrome(int number);
+int GetLength(int number);
+int GetNewNumber(int number);
+
+int main()
+{
+	int number = 0;
+	std::cin >> number;
+	std::cout << (IsPalindrome(number)?"yes":"no");
+}
+
+bool IsPalindrome(int number)
+{
+	int firstEl = 0;
+	int lastEl = 0;
+	int numberCopy = number;
+	while (numberCopy/10>0)
+	{
+		firstEl = numberCopy / pow(10, GetLength(numberCopy) - 1);
+		lastEl = numberCopy % 10;
+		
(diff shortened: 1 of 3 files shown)


SI/Sem.05/Pract.05/Task04.cpp
```diff
@@ -0,0 +1,52 @@
...
+int GetNewNumber(int number)
+{
+	int newNumber = 0;
+
+	for (int i = GetLength(number)-2 ; i > 0; i--)
+	{
+		int currNum = number / pow(10, i);
+		newNumber = newNumber * 10 + currNum % 10;
+	}
+	return newNumber;
```
GeorgiTerziev02, line 51: The point of the task is to convert the number into array and then check if the array is a palindrome

SI/Sem.05/Pract.05/Task05.cpp
```diff
@@ -0,0 +1,72 @@
...
+                }
+                return;
+            }
+        }
+    }
+    if (!haveCommonElements)
+    {
+        std::cout << "No common elements!" << std::endl;
+    }
+}
```
GeorgiTerziev02, line 72: This solution works, but it can be done much more simplier: (hint) with just a while loop and two indexes i and j.

SI/Sem.05/Pract.05/Task06.cpp
```diff
@@ -0,0 +1,114 @@
...
+			for (int j = i+1; j < length; j++)
+			{
+				if (numbers[i]==numbers[j])
+				{
+					return true;
+				}
+			}
+		}
+		return false;
+	}
```
GeorgiTerziev02, line 112: The solution is much simpler than this. Just iterate through the array and if the numbers are not in the correct order - swap them.

GeorgiTerziev02 (changes requested):

PR #160, opened 2022-11-09
Pract. 05: Add 01-03 | Georgi Baltiev

georgibaltiev:
Georgi Baltiev
2 2MI0600168

### DIFF
--- a/Task01.cpp
+++ b/Task01.cpp
@@ -1,0 +1,35 @@
+#include <iostream>
+
+void initializeArray(int a[], int size) {
+    std::cout << "Please enter your numbers:" << std::endl;
+    for (int i = 0; i < size; i++)
+    {
+        std::cin >> a[i];
+    }
+}
+
+bool foundARepetitive(int a[], int size) {
+    const int FIRST_BOUND = size - 1;
+    for (int i = 0; i < FIRST_BOUND; i++)
+    {
+        for (int j = (i + 1); j < size; j++)
+        {
+            if (a[i] == a[j])
+            {
+                std::cout << "The first repeating element is " << a[j] << std::endl;
+                return true;
+            }
+        }
+    }
+    std::cout << "All elements are unique in this array" << std::endl;
+    return false;
+}
+
+int main()
+{
+    const int SIZE = 7;
+    int numbers[SIZE];
+    initializeArray(numbers, SIZE);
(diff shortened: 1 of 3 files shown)


Task01.cpp
```diff
@@ -0,0 +1,35 @@
...
+    {
+        std::cin >> a[i];
+    }
+}
+
+bool foundARepetitive(int a[], int size) {
+    const int FIRST_BOUND = size - 1;
+    for (int i = 0; i < FIRST_BOUND; i++)
+    {
+        for (int j = (i + 1); j < size; j++)
```
GeorgiTerziev02, line 15: No need of brackets

Task01.cpp
```diff
@@ -0,0 +1,35 @@
...
+void initializeArray(int a[], int size) {
+    std::cout << "Please enter your numbers:" << std::endl;
+    for (int i = 0; i < size; i++)
+    {
+        std::cin >> a[i];
+    }
+}
+
+bool foundARepetitive(int a[], int size) {
+    const int FIRST_BOUND = size - 1;
```
GeorgiTerziev02, line 12: No need of this const

Task02.cpp
```diff
@@ -0,0 +1,85 @@
...
+    const int EVEN_AMOUNT = size - ODD_AMOUNT;
+
+    for (int i = 0; i < size - 1; i++)
+    {
+        for (int j = (i + 1); j < size; j++)
+        {
+            if (a[i] % 2 != 0 && a[j] % 2 == 0)
+                swapElements(a[i], a[j]);
+        }
+    }
```
GeorgiTerziev02, line 71: There is a way to place all the even numbers in the beginning with only iteration through the array which is more optimal

Task03.cpp
```diff
@@ -0,0 +1,49 @@
...
+    std::cout << "Please enter your number: " << std::endl;
+    std::cin >> number;
+    if (number <= 0) {
+        std::cout << "Invalid input!" << std::endl;
+        return 0;
+    }
+    else
+    {
+        const int SIZE = findArraySize(number);
+        int binaryArray[SIZE];
```
GeorgiTerziev02, line 44: This is not working

GeorgiTerziev02 (changes requested):

PR #161, opened 2022-11-09
Pract. 05: Add 04-06 | Georgi Baltiev

georgibaltiev:
Georgi Baltiev
2 2MI0600168

### DIFF
--- a/Task04.cpp
+++ b/Task04.cpp
@@ -1,0 +1,59 @@
+#include <iostream>
+
+int returnNumberLength(int number)
+{
+    int i = 0;
+    while (number > 0)
+    {
+        number /= 10;
+        i++;
+    }
+    return i;
+}
+
+void breakDownIntoAnArray(int a[], int size, int number)
+{
+    for (int i = 0; i < size; i++)
+    {
+        a[i] = number % 10;
+        number /= 10;
+    }
+}
+
+bool isPalindrome(int a[], int size)
+{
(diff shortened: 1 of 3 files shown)


Task04.cpp
```diff
@@ -0,0 +1,59 @@
...
+        return 0;
+    }
+
+    else if (number >= 0 && number < 10)
+    {
+
+        std::cout << "One-digit numbers are always palindromes!" << std::endl;
+        return 0;
+    }
+    else
```
GeorgiTerziev02, line 51: no need of these elses

Task04.cpp
```diff
@@ -0,0 +1,59 @@
...
+    {
+        const int SIZE = returnNumberLength(number);
+        int numberHolder[SIZE];
```
GeorgiTerziev02, line 54: You can not initialize array like this

Task06.cpp
```diff
@@ -0,0 +1,48 @@
...
+
+void printArray(int a[], int size) {
+    for(int i = 0; i < size; i++) {
+        std::cout << a[i] << " ";
+    }
+    std::cout << std::endl;
+}
+
+void turnArrayIntoASaw(int a[], int size) {
+    bubbleSort(a, size);
```
GeorgiTerziev02, line 36: No need of the bubble sort

GeorgiTerziev02 (changes requested):

PR #162, opened 2022-11-09
Pract. 05: Add 07-08 | Georgi Baltiev

georgibaltiev:
Georgi Baltiev
2 2MI0600168

### DIFF
--- a/Task08.cpp
+++ b/Task08.cpp
@@ -1,0 +1,57 @@
+#include <iostream>
+
+void printMatrix(int a[25][25], int row, int column) {
+    for(int i = 0; i < row; i++) {
+        for(int j = 0; j < column; j++) {
+            std::cout << a[i][j] << " ";
+        }
+        std::cout << std::endl;
+    }
+}
+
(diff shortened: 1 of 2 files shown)


Task08.cpp
```diff
@@ -0,0 +1,57 @@
...
+void initializeMatrix(int a[25][25], int row, int column) {
+    std::cout << "Please fill in your matrix" << std::endl;
+    for(int i = 0; i < row; i++) {
+        for(int j = 0; j < column; j++){
+        std::cin >> a[i][j];
+    }
+    }
+}
+
+void multiplyAndPrint(int a[25][25], int b[25][25], int rowA, int colA_rowB, int colB) {
```
GeorgiTerziev02, line 21: Note: Single responsibility principle - One function should do one thing => one function should only multiply funcs and one function should only print arrays

GeorgiTerziev02 (changes requested):
Overally: Great

Justsvetoslavov:
Not in correct folder.
Be aware of position of the solutions in terms of specific folder.
Address the comments and contact us if you have questions.

PR #163, opened 2022-11-09
Pract. 05: Add Tic Tac Toe | Georgi Baltiev

georgibaltiev:
Georgi Baltiev 
2 2MI0600168


TicTacToeGame.cpp
```diff
@@ -0,0 +1,97 @@
...
+    for(int i = 0; i < 3; i++) {
+        if(a[i][0] + a[i][1] + a[i][2] == (3 * player)) return true;
+        if(a[0][i] + a[1][i] + a[2][i] == (3 * player)) return true;
+    }
+    if(a[0][0] + a[1][1] + a[2][2] == (3 * player) || a[0][2] + a[1][1] + a[2][0] == (3 * player)) return true;
+    return false;
+}
+
+int main() {
+    int board[3][3];
```
GeorgiTerziev02, line 66: Board should a global variable, because it is the most important part of your program, so you can use it easily in all the functions

TicTacToeGame.cpp
```diff
@@ -0,0 +1,97 @@
+#include <iostream>
+#include <iomanip>
+
+void initializeBoard(int a[3][3]) {
+    for(int i = 0; i < 3; i++) {
+        for(int j = 0; j < 3; j++) {
+            a[i][j] = 0;
+        }
+    }
+}
```
GeorgiTerziev02, line 10: A bit useless function - when you initialize board you can write int board[3][3] = {0}, it will initialize the whole matrix with zeros

TicTacToeGame.cpp
```diff
@@ -0,0 +1,97 @@
...
+            break;
+        } else std::cout << "Sorry, but that space isn't available / doesn't exist! Try again! " << std::endl;
+    } while(isNotAvailable(row, col, a));
+}
+
+char symbol(int a) {
+    if(a == 1) return 'X';
+    else if(a == -1) return 'O';
+    else return ' ';
+}
```
GeorgiTerziev02, line 43: No need of elses here

TicTacToeGame.cpp
```diff
@@ -0,0 +1,97 @@
...
+    }
+    if(a[0][0] + a[1][1] + a[2][2] == (3 * player) || a[0][2] + a[1][1] + a[2][0] == (3 * player)) return true;
+    return false;
+}
+
+int main() {
+    int board[3][3];
+    initializeBoard(board);
+
+    const int PLAYER_X = 1, PLAYER_O = -1;
```
GeorgiTerziev02, line 69: Move constants outside of main

TicTacToeGame.cpp
```diff
@@ -0,0 +1,97 @@
...
+}
+
+char symbol(int a) {
+    if(a == 1) return 'X';
+    else if(a == -1) return 'O';
+    else return ' ';
+}
+
+void printOutBoard(int a[3][3]) {
+        std::cout << "     1   2   3 " << std::endl;
```
GeorgiTerziev02, line 46: Before printing the board clear the console
Use these two cool lines

    std::cout << "\033[;H"; // Moves cursor to the top left
    std::cout << "\033[J"; // Clears the console

TicTacToeGame.cpp
GeorgiTerziev02, line 66: Board can be a char matrix

TicTacToeGame.cpp
GeorgiTerziev02, line 43: If board was a char matrix, you would not have this func

TicTacToeGame.cpp
```diff
@@ -0,0 +1,97 @@
...
+}
+
+void enterASymbol(int a[3][3], int player) {
+    int row, col;
+    do{
+        if(player == 1) std::cout << "It's X turn! Please enter your coordinates: " << std::endl;
+        else std::cout << "It's O turn! Please enter your coordinates: " << std::endl;
+        std::cin >> row >> col;
+        row--;
+        col--;
```
GeorgiTerziev02, line 31: You should tell the user what he is entering first row or col

TicTacToeGame.cpp
```diff
@@ -0,0 +1,97 @@
...
+
+bool isNotAvailable(int row, int col, int a[3][3]) {
+    return (isOutOfBounds(row, col) || isOccupied(row, col, a));
+}
+
+void enterASymbol(int a[3][3], int player) {
+    int row, col;
+    do{
+        if(player == 1) std::cout << "It's X turn! Please enter your coordinates: " << std::endl;
+        else std::cout << "It's O turn! Please enter your coordinates: " << std::endl;
```
GeorgiTerziev02, line 28: Some new lines would have been great

TicTacToeGame.cpp
```diff
@@ -0,0 +1,97 @@
...
+        std::cin >> row >> col;
+        row--;
+        col--;
+        if(!isNotAvailable(row, col, a)) {
+            a[row][col] = player;
+            break;
+        } else std::cout << "Sorry, but that space isn't available / doesn't exist! Try again! " << std::endl;
+    } while(isNotAvailable(row, col, a));
```
GeorgiTerziev02, line 36: Try to optimize this check - You are checking a position two times on every iteration, you can do it with only one check.
While(true) is an option

TicTacToeGame.cpp
```diff
@@ -0,0 +1,97 @@
...
+        } else {
+        enterASymbol(board, PLAYER_O);
+        playerXTurn = true;
+        printOutBoard(board);
+        counter++;
+        if(aPlayerHasWon(board, PLAYER_O)) {
+            std::cout << "A victory for O!" << std::endl;
+            return 0;
+        } 
+        }
```
GeorgiTerziev02, line 93: Lots of duplicate code here

TicTacToeGame.cpp
```diff
@@ -0,0 +1,97 @@
...
+    if(a[0][0] + a[1][1] + a[2][2] == (3 * player) || a[0][2] + a[1][1] + a[2][0] == (3 * player)) return true;
+    return false;
+}
+
+int main() {
+    int board[3][3];
+    initializeBoard(board);
+
+    const int PLAYER_X = 1, PLAYER_O = -1;
+    bool playerXTurn = true;
```
GeorgiTerziev02, line 70: You can make this a global variable

TicTacToeGame.cpp
```diff
@@ -0,0 +1,97 @@
...
+    printOutBoard(board);
+    int counter = 0;
+
+    while(counter < 9){
```
GeorgiTerziev02, line 74: No need of the counter, the game ends only if a user has won

georgibaltiev: Yeah, but what about when they are tied? I tried it once and the program got locked in a stalemate without being able to exit.

GeorgiTerziev02: My mistake, didn't think of it. Sorry.

GeorgiTerziev02 (approved):

PR #166, opened 2022-11-10
Pract.05: Add 04-06 | Silvana Nikolova

SilvanaNikolova:
Silvana Nikolova
2 0MI0600152

### DIFF
--- a/SI/Sem.05/Pract.05/Task04.cpp
+++ b/SI/Sem.05/Pract.05/Task04.cpp
@@ -1,0 +1,32 @@
+﻿#include <iostream>
+using namespace std;
+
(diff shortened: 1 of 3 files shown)


SI/Sem.05/Pract.05/Task04.cpp
```diff
@@ -0,0 +1,32 @@
...
+				resultArr[counter++] = currEl;
+			}
+		}
+	}
+
+	for (int i = 0; i < counter; ++i)
+	{
+		cout << resultArr[i] << ' ';
+	}
+}
```
GeorgiTerziev02, line 24: which task is this?

GeorgiTerziev02 (changes requested):

PR #167, opened 2022-11-10
Pract.05: Add 07-08 | Silvana Nikolova

SilvanaNikolova:
Silvana Nikolova
2 0MI0600152

### DIFF
--- a/SI/Sem.05/Pract.05/Task07.cpp
+++ b/SI/Sem.05/Pract.05/Task07.cpp
@@ -1,0 +1,35 @@
+#include <iostream>
+using namespace std;
+
+void transformMatrix(int** matrix, int size, int** resultMatrix)
+{
+	for (int i = 0; i < size; ++i)
+	{
+		for (int j = 0; j < size; ++j)
+		{
+			resultMatrix[j][i] = matrix[i][j];
+		}
+	}
+
+	for (int i = 0; i < size; ++i)
+	{
+		for (int j = 0; j < size; ++j)
+		{
+			cout << resultMatrix[i][j] << ' ';
+		}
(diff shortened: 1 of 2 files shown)


SI/Sem.05/Pract.05/Task08.cpp
```diff
@@ -0,0 +1,45 @@
+#include <iostream>
+using namespace std;
+
+void multiplyMatrix(int **firstMatrix, int **secondMatrix, int **resultMatrix, int size)
```
GeorgiTerziev02, line 4: You should pass as parameters the dimensions of the arrays

SI/Sem.05/Pract.05/Task08.cpp
```diff
@@ -0,0 +1,45 @@
...
+	int firstEl = 0;
+	int secondEl = 0;
+
+	for (int row = 0; row < 3; row++)
+	{
+		for (int col = 0; col < 3; col++)
+		{
+			for (int j = 0; j < 3; j++)
+			{
+				firstEl = firstMatrix[row][col];
```
GeorgiTerziev02, line 17: matrix[row][j]

SI/Sem.05/Pract.05/Task07.cpp
```diff
@@ -0,0 +1,35 @@
+#include <iostream>
+using namespace std;
+
+void transformMatrix(int** matrix, int size, int** resultMatrix)
```
GeorgiTerziev02, line 4: You should pass the matrixes as int matrix[][some very big number]

GeorgiTerziev02 (changes requested):

PR #171, opened 2022-11-18
Pract.06: Add 02-03-05 | Ivan Stoynev

Vacata12:
Ivan Stoynev
3 6MI0700164

### DIFF
--- a/IS/Sem.06/Pract.06/Task03.cpp
+++ b/IS/Sem.06/Pract.06/Task03.cpp
@@ -1,0 +1,22 @@
+#include <iostream>
+using namespace std;
+
+int theSpecByte(int num, int bit)
+{
+	int count = 0, result = 0;
+	while (count != bit)
+	{
+		(num % 2 != 0) ? result = 1 : result = 0;
+		num /= 2;
+		count++;
+	}
+	return result;
+}
+
+int main()
+{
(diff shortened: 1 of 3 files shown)


IS/Sem.06/Pract.06/Task03.cpp
```diff
@@ -0,0 +1,22 @@
...
+{
+	int count = 0, result = 0;
+	while (count != bit)
+	{
+		(num % 2 != 0) ? result = 1 : result = 0;
+		num /= 2;
+		count++;
+	}
+	return result;
+}
```
GeorgiTerziev02, line 14: The point of the task is to use bitwise operations

IS/Sem.06/Pract.06/Task05.cpp
```diff
@@ -0,0 +1,40 @@
...
+		}
+		else
+		{
+			result += powOfTwo(copyOfK - k);
+			num /= 2;
+			--k;
+		}
+	}
+	return result;
+}
```
GeorgiTerziev02, line 32: The point of the task is to use bitwise operations
Think of a better name for this function

GeorgiTerziev02 (changes requested):
Use bitwise operations in task 3 and 5

PR #179, opened 2022-11-23
Pract.06: Add 04 | Daniel Jordanov

danielgo26:
Daniel Jordanov
6 3MI0600149


Justsvetoslavov (changes requested):
The solution is TOO complex, because you are NOT using bitwise operators.
Revise this task : )

PR #181, opened 2022-11-28
Sem 07 tasks

GeorgiTerziev02:
I can optimize the tasks not to use strLen if needed.


IS/Sem.07/Solutions.cpp
```diff
@@ -0,0 +1,135 @@
...
+
+    while (str1[i] == str2[i]) {
+        if (str1[i] == '\0'/* && str2[i] == '\0'*/) {
+            return 0;
+        }
+
+        i++;
+    }
+
+    return str1[i] - str2[i] > 0 ? 1 : -1;
```
Justsvetoslavov, line 24: Ternary is not necessary.
Just return str1[i] - str2[i]

IS/Sem.07/Solutions.cpp
```diff
@@ -0,0 +1,135 @@
...
+}
+
+void myStrCat(const char* str1, const char* str2, char* result) {
```
Justsvetoslavov, line 27: We can use only dest and source.

Suggestion:

```
void my_strcat(char* dest, const char* source)
{
	size_t destLength = my_strlen(dest);
	size_t sourceLength = my_strlen(source);

	for (size_t i = 0; i < sourceLength; i++)
	{
		dest[destLength + i] = source[i];
	}
	dest[destLength + sourceLength] = '\0';
}
```

IS/Sem.07/Solutions.cpp
```diff
@@ -0,0 +1,135 @@
...
+
+    for (size_t i = 0; i < str2Length; i++) {
+        result[str1Length + i] = str2[i];
+    }
+
+    result[str1Length + str2Length] = '\0';
+}
+
+
+void myStrCopy(char* destination, const char* source) {
```
Justsvetoslavov, line 43: rename to dest (not necessary, as it is a preference)

Justsvetoslavov (commented):
Should we consider splitting into multiple .cpps ,as they are different tasks?

IS/Sem.07/Solutions.cpp
GeorgiTerziev02: The students wanted it with -1 and 1 :D

Justsvetoslavov (approved):
👏

PR #187, opened 2022-11-30
Prac07 - Strings: Add | Task04 - Presiyana

PresiyanaBB:
Presiyana 3MI0600240 6

### DIFF
--- a/SI/Sem.07/Pract.07/Strings/Task04.cpp
+++ b/SI/Sem.07/Pract.07/Strings/Task04.cpp
@@ -20,7 +20,7 @@ void replaceSymbols(char* arr, char x, char a, char b)
 
 int main()
 {
-	char arr[] = "H_ll_ my fri_nd\0";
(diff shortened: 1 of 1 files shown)


Justsvetoslavov (approved):

SI/Sem.07/Pract.07/Strings/Task04.cpp
```diff
@@ -0,0 +1,27 @@
+#include <iostream>
+using namespace std;
+
+void replaceSymbols(char* arr, char x, char a, char b)
```
dejansvetoslavov, line 4: Parameters should be named better

Justsvetoslavov: @PresiyanaBB could you address the naming comment?

PresiyanaBB:
They are named this way in the task that is why I used them like that

На пн, 5.12.2022 г., 10:58 ч. Nikola Svetoslavov ***@***.***>
написа:

PR #189, opened 2022-11-30
Pract.07: Add 07 | Elis Shukri

Eliss12:
Elis Shukri
6 8MI0600194

### DIFF
--- a/SI/Sem.07/Pract.07/Strings/Task07.cpp
+++ b/SI/Sem.07/Pract.07/Strings/Task07.cpp
@@ -1,19 +1,19 @@
 ﻿
 
 #include <iostream>
-const int SIZE = 50;
-int compareStrings(const char firstString[], const char secondString[])
+const int SIZE = 1024;
+int compareStrings(const char* firstString, const char* secondString)
 {
-	int index = 0;
-	while (firstString[index] == secondString[index]) 
+	size_t index = 0;
+	while (firstString[index] == secondString[index])
 	{
-		if (firstString[index] == '\0') 
+		if (firstString[index] == '\0')
 		{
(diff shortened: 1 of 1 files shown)


SI/Sem.07/Pract.07/Strings/Task07.cpp
```diff
@@ -0,0 +1,31 @@
...
+	size_t index = 0;
+	while (firstString[index] == secondString[index])
+	{
+		if (firstString[index] == '\0')
+		{
+			return 0;
+		}
+		index++;
+	}
+	if (firstString[index] - secondString[index] < 0)
```
Justsvetoslavov, line 16: we can use ternary for this if-else statement.

return + ternary.

SI/Sem.07/Pract.07/Strings/Task07.cpp
```diff
@@ -0,0 +1,31 @@
+﻿
```
Justsvetoslavov, line 1: remove first 2 empty lines

SI/Sem.07/Pract.07/Strings/Task07.cpp
```diff
@@ -0,0 +1,31 @@
...
+			return 0;
+		}
+		index++;
+	}
+	if (firstString[index] - secondString[index] < 0)
+	{
+		return -1;
+	}
+	else { return 1; }
+}
```
Justsvetoslavov, line 21: 1 newline for separator between the 2 functions

Justsvetoslavov (changes requested):

Justsvetoslavov (approved):

PR #190, opened 2022-11-30
Pract.07: Add 11 | Sasho Vasilchev

SashoVas:
Sasho Vasilchev
6 0MI0600197


SI/Sem.07/Pract.07/Strings/Task11.cpp
```diff
@@ -0,0 +1,51 @@
...
+}
+void Replace(char* str, int start, int end) {
+    for (int i = start; i < end; i++)
+    {
+        str[i] = '*';
+    }
+}
+int main()
+{
+    int UpperCaseDifferance = 'A' - 'a';
```
Justsvetoslavov, line 21: Can we split the code in main to specific function/s?

SI/Sem.07/Pract.07/Strings/Task11.cpp
```diff
@@ -0,0 +1,55 @@
+#include <iostream>
+using namespace std;
+const int MaxLength = 1000;
+const int UpperCaseDifferance = 'A' - 'a';
+int GetLenght(char* str) {
```
Justsvetoslavov, line 5: const char*

SI/Sem.07/Pract.07/Strings/Task11.cpp
```diff
@@ -0,0 +1,55 @@
...
+    int length = 0;
+    while (*str)
```
Justsvetoslavov, line 7: I think:
```
    while (str[index] != '\0') {
        index++;
    }
```
is more readable but your call.

Justsvetoslavov (commented):
I think the idea of the task was to match exact words, rather than every substring, thus I would suggest to:
- Add a Task11 notes that are specific to this solution.
- Rename the current task11 notes to task12.

If you have any concerns , feel free to ping me (in facebook or here)

Justsvetoslavov (approved):

PR #191, opened 2022-11-30
Pract07: Add 02 | Militsa Toncheva

milicavilica:
Militsa Toncheva
Group 6 2MI0600216

### DIFF
--- a/SI/Sem.07/Pract.07/Strings/Task02.cpp
+++ b/SI/Sem.07/Pract.07/Strings/Task02.cpp
@@ -1,0 +1,53 @@
+#include <iostream>;
+
+using namespace std;
+
+const int SIZE = 1000;
+bool CheckInput(int a)
+{
+	return a > 0;
+}
+int GetStringLenght(char* str)
+{
+	int i = 0;
+	while (str[i] != '\0')
+	{
+		i++;
+	}
+	return i;
+}
+int NewString(char* str, int beg, int end)
+{
+	int strLenght = GetStringLenght(str);
+	if (beg > strLenght)
+	{
+		cout << " ";
+		return 0;
+	}
+	if (end > strLenght)
+	{
+		while (str[beg] != '\0')
+		{
+			cout << str[beg];
+			beg++;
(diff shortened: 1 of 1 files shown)


SI/Sem.07/Pract.07/Strings/Task02.cpp
```diff
@@ -0,0 +1,53 @@
+#include <iostream>;
+
+using namespace std;
+
+const int SIZE = 1000;
+bool CheckInput(int a)
+{
+	return a > 0;
```
Justsvetoslavov, line 8: can beg be 0 ?
Isn't 0 a valid index?

SI/Sem.07/Pract.07/Strings/Task02.cpp
```diff
@@ -0,0 +1,53 @@
...
+}
+int GetStringLenght(char* str)
```
Justsvetoslavov, line 10: const char*

SI/Sem.07/Pract.07/Strings/Task02.cpp
```diff
@@ -0,0 +1,53 @@
...
+	if (end > strLenght)
+	{
+		while (str[beg] != '\0')
+		{
+			cout << str[beg];
+			beg++;
+		}
+		return 0;
+	}
+	for (int i = beg; i <= end; i++)
```
Justsvetoslavov, line 36: fix format

SI/Sem.07/Pract.07/Strings/Task02.cpp
```diff
@@ -0,0 +1,53 @@
...
+int GetStringLenght(char* str)
+{
+	int i = 0;
+	while (str[i] != '\0')
+	{
+		i++;
+	}
+	return i;
+}
+int NewString(char* str, int beg, int end)
```
Justsvetoslavov, line 19: Should this function be an int?

Justsvetoslavov (changes requested):

PR #193, opened 2022-11-30
Pract.07: References Add 07 | Daniel Jordanov

danielgo26:
Daniel Jordanov
6 3MI0600149

### DIFF
--- a/SI/Sem.07/Pract.07/References/Task07.cpp
+++ b/SI/Sem.07/Pract.07/References/Task07.cpp
@@ -26,9 +26,7 @@ bool GetInput(int movies[MAX_COUNT_ROWS][MAX_COUNT_COLS], int countRows)
 {
 	for (int i = 0; i < countRows; i++)
 	{
-		int hourBegin = 0;
-		int minutesBegin = 0;
-		int duration = 0;
+		int hourBegin, minutesBegin, duration;
 		std::cin >> hourBegin >> minutesBegin >> duration;
 		if (!CheckValidInput(hourBegin, minutesBegin, duration))
 		{
@@ -37,7 +35,7 @@ bool GetInput(int movies[MAX_COUNT_ROWS][MAX_COUNT_COLS], int countRows)
 		}
 		else
 		{
-			
+
 			movies[i][0] = hourBegin;
 			movies[i][1] = minutesBegin;
 			movies[i][2] = duration;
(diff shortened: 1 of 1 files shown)


SI/Sem.07/Pract.07/References/Task07.cpp
```diff
@@ -0,0 +1,125 @@
...
+int* SumTime(int hoursBegin, int minutesBegin, int hoursDuration, int minutesDuration)
+{
+	int endTime[2];
+	int hoursEnd = hoursBegin + hoursDuration;
+	int minutesEnd = minutesBegin + minutesDuration;
+
+	endTime[0] = hoursEnd;
+	endTime[1] = minutesEnd;
+	return endTime;
+}
```
GeorgiTerziev02, line 102: What if minutesBegin + minutesDuration >= 60? Also this function again won't work as you expect it to work. The reason is the same;

SI/Sem.07/Pract.07/References/Task07.cpp
```diff
@@ -0,0 +1,125 @@
...
+	int maxLatestMinutes = -1;
+
+	for (int i = 0; i < countRows; i++)
+	{
+		int duration = movies[i][2];
+		int hoursOfDuration = ConvertToMinutes(duration)[0];
+		int minutesOfDuration = ConvertToMinutes(duration)[1];
+
+		int endHour = SumTime(movies[i][0], movies[i][1], hoursOfDuration, minutesOfDuration)[0];
+		int endMinutes = SumTime(movies[i][0], movies[i][1], hoursOfDuration, minutesOfDuration)[1];
```
GeorgiTerziev02, line 70: @Justsvetoslavov Why these two lines work correctly?

GeorgiTerziev02: + you are calculating the same thing two times

SI/Sem.07/Pract.07/References/Task07.cpp
```diff
@@ -0,0 +1,125 @@
...
+	endTime[0] = hoursEnd;
+	endTime[1] = minutesEnd;
+	return endTime;
+}
+
+void Print(int hoursEnd, int minutesEnd)
+{
+	if (hoursEnd >= 24)
+	{
+		hoursEnd -= (hoursEnd / 24) * 24;
```
GeorgiTerziev02, line 108: %= 24

GeorgiTerziev02 (changes requested):

SI/Sem.07/Pract.07/References/Task07.cpp
Justsvetoslavov: @GeorgiTerziev02 We have reworked the solution in class.

As for the above question, it is returning a pointer and we are referencing by offset afterwards.(really hacky which we removed)

Justsvetoslavov (approved):

PR #194, opened 2022-11-30
Pract.07: Add 09 | Mihail Yordanov

BojeMoii:


Justsvetoslavov:
PR description is missing

SI/Task09.cpp
```diff
@@ -0,0 +1,47 @@
+#include <iostream>
+
+int Broqch(char* arr)
```
GeorgiTerziev02, line 3: Change the name of the function, const char* arr - you are not modifying the char[]

SI/Task09.cpp
```diff
@@ -0,0 +1,47 @@
...
+
+int Index(char* text, char* word)
+{
+	int pom = -1;
+	for (int i = 0; i < Broqch(text); i++)
+	{
+		if (text[i] == word[0])
+		{
+			bool isTrue = true;
+			for (int j = 0; j < Broqch(word); j++)
```
GeorgiTerziev02, line 21: Big mistake - fix the for loop - you are calculating the length of the char[] **after every iteration** of the for loop
=> size_t length = myStrLen(word) and then use length in the condition of the for loop

SI/Task09.cpp
```diff
@@ -0,0 +1,47 @@
...
+			{
+				if (text[i + j] != word[j])
+				{
+					isTrue = false;
+				}
```
GeorgiTerziev02, line 26: What if i + j is bigger or equal to the length of the text?

SI/Task09.cpp
```diff
@@ -0,0 +1,47 @@
...
+	{
+		br++;
+	}
+	return br;
+}
+
+int Index(char* text, char* word)
+{
+	int pom = -1;
+	for (int i = 0; i < Broqch(text); i++)
```
GeorgiTerziev02, line 16: Big mistake - fix the for loop - you are calculating the length of the char[] after every iteration of the for loop
=> size_t length = myStrLen(text) and then use length in the condition of the for loop

GeorgiTerziev02 (changes requested):

PR #195, opened 2022-11-30
Pract.07: Add task 09 | Krasimir Ronkov

Krasimir-ronkov:


Justsvetoslavov:
PR description is missing

SI/Sem.07/Pract.07/References/task09.cpp
```diff
@@ -0,0 +1,47 @@
...
+    int b_copy = b;
+    int count = 0;
+    while(b_copy!=0) {
+        count++;
+        b_copy /= 10;
+    }
+    int pow = 1;
+    for (int i = 0; i < count;i++) {
+        pow *= 10;
+    }
```
GeorgiTerziev02, line 33: You can calculate pow in the while loop above, the for is not needed

GeorgiTerziev02 (approved):
You can have better names for your variables, also you can create some functions so that your code can be more readable

Justsvetoslavov (changes requested):
Address @GeorgiTerziev02 comment.
Remove the comments.
Move the code to appropriate functions.

PR #197, opened 2022-11-30
Pract.07: Add 03-04 | Danail Todorov

Danail3:
Danail Todorov
6 0MI0600241


Justsvetoslavov:
Edit title , does not match PR rules.

SI/Sem.07/Pract.07/References/Pract07-04.cpp
```diff
@@ -0,0 +1,55 @@
...
+	SwapValuesIfHigher(lower, higher);
+	for (int i = 1; i <= higher; i++)
+	{
+		if (lower % i == 0 && higher % i == 0)
+		{
+			gcd = i;
+		}
+	}
+	return gcd;
+}
```
GeorgiTerziev02, line 26: There are much better algorithms for gcd, think what you have leant in the workshops.

SI/Sem.07/Pract.07/References/Pract07-04.cpp
```diff
@@ -0,0 +1,55 @@
...
+	int secondGCD = FindGCD(a, b);
+	n /= firstGCD;
+	m /= firstGCD;
+	a /= secondGCD;
+	b /= secondGCD;
+}
+
+void SumOfFractions(int& a, int& b, int& c, int& d)
+{
+	Reduction(a, b, c, d);
```
GeorgiTerziev02, line 40: You should call reduction two times, once for a and b, once for c and d

SI/Sem.07/Pract.07/References/Pract07-04.cpp
```diff
@@ -0,0 +1,55 @@
...
+	int gcd = FindGCD(b, d);
+	b *= (d / gcd);
+	a *= (d / gcd);
+	d *= (b / gcd);
+	c *= (b / gcd);
```
GeorgiTerziev02, line 45: You repeat some calculations (b/gcd => two times)

GeorgiTerziev02: Your code can be cleaner and more readble like this
int numerator = a*d + c*b, denominator = b*d;
Reduction(numerator, denominator)
then cout
instead of these calculations which are not obvious at first sight

PR #199, opened 2022-12-01
Pract. 07 (References): Add 01-03 | Georgi Baltiev

georgibaltiev:
Georgi Baltiev
2 2MI0600168

### DIFF
--- a/SI/Task03.cpp
+++ b/SI/Task03.cpp
@@ -1,0 +1,37 @@
+#include <iostream>
+
+void sort(int& min, int& mid, int& max) {
+    int a = min, b = mid, c = max;
+    if(a <= b && b <= c) {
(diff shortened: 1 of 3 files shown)


SI/Task03.cpp
```diff
@@ -0,0 +1,37 @@
...
+    } else if (c <= a && a <= b) {
+        min = c;
+        mid = a;
+        max = b;
+    } else {
+        min = c;
+        mid = b;
+        max = a;
+    }
+}
```
GeorgiTerziev02, line 30: Can be optimized
if a < b => swap
then if b < c => swap
then if a < b => swap

GeorgiTerziev02 (changes requested):

PR #200, opened 2022-12-01
Pract. 07 (References): Add 04-06 | Georgi Baltiev

georgibaltiev:
Georgi Baltiev
2 2MI0600168

### DIFF
--- a/Task04.cpp
+++ b/Task04.cpp
@@ -1,0 +1,55 @@
+#include <iostream>
+
+int GCD(int alpha, int beta) {
+    if(alpha == beta) return alpha;
+    int first = (alpha > beta) ? alpha : beta;
+    int second = (alpha < beta) ? alpha : beta;
+        do{
+             int remainder = first % second;
+             first = second;
+             second = remainder;
+        } while (first != 0);
+        return second;
+}
+
+void simplifyFraction(int& numerator, int& denominator) {
+    if(numerator == denominator) {
+        numerator = 1;
+        denominator = 1;
+    } else {
+    int commonDivisor = GCD(numerator, denominator); 
+    while(commonDivisor != 1) {
+        numerator /= commonDivisor;
+        denominator /= commonDivisor;
+        commonDivisor = GCD(numerator, denominator);
+    }
+}
+}
+
+void printOutFraction(int num, int denom) {
+    std::cout << num << "/" << denom;
+}
+
+void printOutTheEquation(int firstNumerator, int firstDenominator, int secondNumerator, int secondDenominator, int sumNum, int sumDenom) {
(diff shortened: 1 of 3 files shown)


Task04.cpp
```diff
@@ -0,0 +1,55 @@
+#include <iostream>
+
+int GCD(int alpha, int beta) {
+    if(alpha == beta) return alpha;
+    int first = (alpha > beta) ? alpha : beta;
+    int second = (alpha < beta) ? alpha : beta;
```
GeorgiTerziev02, line 6: No needed, just use alpha and beta, try it out and think why it is correct

Task04.cpp
```diff
@@ -0,0 +1,55 @@
...
+    if(numerator == denominator) {
+        numerator = 1;
+        denominator = 1;
+    } else {
+    int commonDivisor = GCD(numerator, denominator); 
+    while(commonDivisor != 1) {
+        numerator /= commonDivisor;
+        denominator /= commonDivisor;
+        commonDivisor = GCD(numerator, denominator);
+    }
```
GeorgiTerziev02, line 25: This is gcd, you should just do, there won't be any other common divisors different than zero :)
        numerator /= commonDivisor;
        denominator /= commonDivisor;

Task05.cpp
```diff
@@ -0,0 +1,24 @@
...
+        powerTen *= 10;
+    }
+    int digitOfFirst = (first / powerTen) % 10;
+    int digitOfSecond = (second / powerTen % 10);
+    if(powerTen <= first && powerTen <= second) {
+    first -= (digitOfFirst * powerTen);
+    second -= (digitOfSecond * powerTen);
+    first += (digitOfSecond * powerTen);
+    second += (digitOfFirst * powerTen);
+    }
```
GeorgiTerziev02, line 15: format your code better, ctrl + k + d

Task05.cpp
```diff
@@ -0,0 +1,24 @@
+#include <iostream>
+
+void switchDigits(int& first, int& second, int k) {
+    int powerTen = 1;
+    for(int i = 1; i < k; i++) {
+        powerTen *= 10;
+    }
+    int digitOfFirst = (first / powerTen) % 10;
+    int digitOfSecond = (second / powerTen % 10);
```
GeorgiTerziev02, line 9: Change position of the right bracket

Task05.cpp
```diff
@@ -0,0 +1,24 @@
...
+    if(powerTen <= first && powerTen <= second) {
```
GeorgiTerziev02, line 10: This check should be done before line 8 and 9

GeorgiTerziev02 (changes requested):

PR #201, opened 2022-12-01
Pract. 07 (References): Add 07-09 | Georgi Baltiev

georgibaltiev:
Georgi Baltiev
2 2MI0600168

### DIFF
--- a/Task07.cpp
+++ b/Task07.cpp
@@ -11,8 +11,14 @@ void calculateHours(int& hour, int& minute, int startHour, int startMinute, int 
         hour = startHour;
         minute = startMinute;
     }
-    else if(startHour == hour) {
-        if(startMinute > minute) minute = startMinute;
+    else if(startHour == hour && minute > startMinute) {
+        minute = startMinute;
+    }
+}
+
+void convertHoursToNormalTime(int& hours) {
+    while(hours >= 24) {
+        hours-= 24;
     }
 }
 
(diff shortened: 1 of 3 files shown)


Task07.cpp
```diff
@@ -0,0 +1,29 @@
...
+int main() {
+    int n, hour = 0, minute = 0;
+    std::cin >> n;
+    if(n < 0) return 0;
+    for(int i = 0; i < n; i++) {
+        int startingHour, startingMinute, duration;
+        std::cin >> startingHour >> startingMinute >> duration;
+        calculateHours(hour, minute, startingHour, startingMinute, duration);
+    }
+    std::cout << hour << " " << minute << std::endl;
```
GeorgiTerziev02, line 28: You do not check if hour is >= 24 when printing
Try the following input
4
19 15 120
20 20 25
18 25 230
21 10 3232

Task07.cpp
```diff
@@ -0,0 +1,29 @@
...
+        startHour++;
+    }
+
+    if(startHour > hour) {
+        hour = startHour;
+        minute = startMinute;
+    }
+    else if(startHour == hour) {
+        if(startMinute > minute) minute = startMinute;
+    }
```
GeorgiTerziev02, line 16: else if(.... && .....)

GeorgiTerziev02 (changes requested):

GeorgiTerziev02 (approved):

Justsvetoslavov:
@GeorgiTerziev02 @georgibaltiev Is it me or the task are not in the correct directory?

Justsvetoslavov (changes requested):
Waiting for a reply on the above comment

PR #203, opened 2022-12-04
Pract.07/Ref : Add 01/02/04 | Ivan Stoynev

Vacata12:
Ivan Stoynev
3 6MI0700164

### DIFF
--- a/IS/Sem.07/Pract.07/References/Task02.cpp
+++ b/IS/Sem.07/Pract.07/References/Task02.cpp
@@ -1,0 +1,44 @@
+#include <iostream>
+using namespace std;
+
+bool isLetter(char symbol);
+void ToLower(char& symbol);
+void ToUpper(char& symbol);
+bool isUpper(char symbol);
+
+int main()
+{
+  char symbol;
+  cin >> symbol;
+  if(isLetter(symbol))
+  {
+    if(isUpper)
+    {
+        ToLower(symbol);
+    }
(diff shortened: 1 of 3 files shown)


IS/Sem.07/Pract.07/References/Task02.cpp
```diff
@@ -0,0 +1,44 @@
...
+}
+
+void ToLower(char& symbol)
+{
+	symbol += 32;	
+}
+
+void ToUpper(char& symbol)
+{
+	symbol -= 32;
```
GeorgiTerziev02, line 43: You are not consistent in naming you functions, the first two are camelCase and the other two are PascalCase.

IS/Sem.07/Pract.07/References/Task04.cpp
```diff
@@ -0,0 +1,46 @@
...
+		for (int i = 2; i <= helper / 2; i++)
+		{
+			if (num1 % i == 0 && num2 % i == 0)
+			{
+				num1 /= i;
+				num2 /= i;
+			}
+		}
+	}
+}
```
GeorgiTerziev02, line 29: Infinite loop, rethink you logic, try to do it without the helper and try to do it with gcd or lcm(lowest common divisor)
Try input 2 2 3 3

GeorgiTerziev02 (changes requested):

PR #205, opened 2022-12-06
Pract. 7 (Strings): Add 01-03 | Georgi Baltiev

georgibaltiev:
Georgi Baltiev
2 2MI0600168


Task01.cpp
```diff
@@ -0,0 +1,20 @@
+#include <iostream>
+
+void alterText(char text[1024], char textCopy[1024]) {
```
plamena-ilieva, line 3: text should be const because you never change it

Task01.cpp
```diff
@@ -0,0 +1,20 @@
...
+    int index = 0, copyIndex = 0;
+    while(text[index] != '\0') {
+        if(text[index] >= 97 && text[index] <= 122) {
```
plamena-ilieva, line 6: you can use characters instead

Task02.cpp
```diff
@@ -0,0 +1,20 @@
+#include <iostream>
+
+void printOutText(int beg, int end, char text[1024]) {
+    for(int i = beg; i <= end; i++) {
+        if(text[i] == '\0') break;
+        std::cout << text[i];
+    }
+    std::cout << std::endl;
```
plamena-ilieva, line 8: you have to return a new string (add it in the arguments list), not print it

Task02.cpp
```diff
@@ -0,0 +1,20 @@
+#include <iostream>
+
+void printOutText(int beg, int end, char text[1024]) {
```
plamena-ilieva, line 3: text should be const

Task02.cpp
```diff
@@ -0,0 +1,20 @@
...
+        if(text[i] == '\0') break;
+        std::cout << text[i];
+    }
+    std::cout << std::endl;
+}
+
+int main() {
+    int beg, end;
+    std::cin >> beg >> end;
+    if(beg > end) std::cout << "";
```
plamena-ilieva, line 14: what if both beg and end are > the length of the string?

plamena-ilieva (changes requested):

PR #206, opened 2022-12-06
Pract. 7 (Strings): Add 04-06 | Georgi Baltiev

georgibaltiev:
Georgi Baltiev
2 2MI0600168

### DIFF
--- a/Task05.cpp
+++ b/Task05.cpp
@@ -1,0 +1,16 @@
+#include <iostream>
+
+int countOccurences(char text[1024], char letter) {
+    int index = 0, counter = 0;
+    while(text[index] != '\0') {
+        if(text[index] == letter) counter++;
+        index++;
+    }
+    return counter;
+}
+
+int main() {
+    char text[1024], letter;
+    std::cin >> text >> letter;
+    std::cout << countOccurences(text, letter) << std::endl;
+}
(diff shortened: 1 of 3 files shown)


Task05.cpp
```diff
@@ -0,0 +1,16 @@
+#include <iostream>
+
+int countOccurences(char text[1024], char letter) {
```
plamena-ilieva, line 3: text should be const

Task06.cpp
```diff
@@ -0,0 +1,18 @@
+#include <iostream>
+
+void printOutLongestPrefix(char text1[1024], char text2[1024], char text3[1024]) {
```
plamena-ilieva, line 3: const

Task06.cpp
```diff
@@ -0,0 +1,18 @@
...
+    int index = 0;
+    while(text1[index] != '\0' && text2[index] != '\0' && text3[index] != '\0') {
+        if(text1[index] == text2[index] && text2[index] == text3[index]) {
+            std::cout << text1[index];
+        } else break;
+        index++;
+    }
+    std::cout << std::endl;
```
plamena-ilieva, line 11: new string, not print

plamena-ilieva (changes requested):

PR #207, opened 2022-12-06
Pract. 7 (Strings): Add 07-09 | Georgi Baltiev

georgibaltiev:
Georgi Baltiev
2 2MI0600168

### DIFF
--- a/Task07.cpp
+++ b/Task07.cpp
@@ -1,0 +1,25 @@
+#include <iostream>
+
+int lexicographicalComparison(char text1[1024], char text2[1024]) {
+    int index = 0;
+    while(text1[index] != 0 && text2[index] != 0) {
+        if(text1[index] > text2[index]) {
+            return 1;
+            break;
+        }
+        if(text1[index] < text2[index]) {
+            return -1;
+            break;
+        }
+        index++;
+    }
+    if(text1[index] != '\0') return 1;
+    if(text2[index] != '\0') return -1;
+    return 0;
+}
+
+int main() {
+    char text1[1024], text2[1024];
+    std::cin >> text1 >> text2;
+    std::cout << lexicographicalComparison(text1, text2);
+}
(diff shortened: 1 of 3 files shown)


Task07.cpp
```diff
@@ -0,0 +1,25 @@
+#include <iostream>
+
+int lexicographicalComparison(char text1[1024], char text2[1024]) {
```
plamena-ilieva, line 3: const

Task08.cpp
```diff
@@ -0,0 +1,28 @@
+#include <iostream>
+
+void printOutBasis(char text[1024]) {
```
plamena-ilieva, line 3: edit the existing string

Task08.cpp
```diff
@@ -0,0 +1,28 @@
...
+    int index = 0;
+    if(text[index] == '\0') std::cout << "" << std::endl;
+    else {
+    while(text[index+1] != '\0') {
+        int paralellIndex = index + 1;
+        while(text[paralellIndex] != '\0') {
```
plamena-ilieva, line 9: it would be better to create a new function for this cycle while

Task09.cpp
```diff
@@ -0,0 +1,25 @@
+#include <iostream>
+
+bool equals(char text[1024], char subText[1024], int start, int length) {
```
plamena-ilieva, line 3: const

Task09.cpp
```diff
@@ -0,0 +1,25 @@
...
+        for(int i = 0; i < length; i++) {
+            if(text[start + i] != subText[i]) return false;
+        }
+        return true;
+}
+
+int count(char text[1024], char subText[1024]) {
+    int subLength = 0;
+    while(subText[subLength] != '\0') subLength++;
```
plamena-ilieva, line 12: create a new function for the length of the string

PR #208, opened 2022-12-06
Pract. 7 (Strings): Add 10-11 | Georgi Baltiev

georgibaltiev:
Georgi Baltiev
2 2MI0600168

### DIFF
--- a/Task11.cpp
+++ b/Task11.cpp
@@ -1,0 +1,34 @@
+#include <iostream>
+
+void censor(char text[1024], int begin, int end) {
+    for(int i = begin; i <a end; i++) {
+        text[i] = '*';
+    }
+}
+
+bool equals(char text[1024], char subText[1024], int start, int length) {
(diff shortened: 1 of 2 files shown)


Task11.cpp
```diff
@@ -0,0 +1,34 @@
+#include <iostream>
+
+void censor(char text[1024], int begin, int end) {
+    for(int i = begin; i <a end; i++) {
```
plamena-ilieva, line 4: typo

Task11.cpp
```diff
@@ -0,0 +1,34 @@
...
+        text[i] = '*';
+    }
+}
+
+bool equals(char text[1024], char subText[1024], int start, int length) {
+        for(int i = 0; i < length; i++) {
```
plamena-ilieva, line 10: const

plamena-ilieva (changes requested):

PR #216, opened 2022-12-29
Pract.08: Add 04-05 | Ivan Stoynev

Vacata12:
Ivan Stoynev
3 6MI0700164


IS/Sem.08/Task05.cpp
```diff
@@ -0,0 +1,97 @@
+#include <iostream>
```
Justsvetoslavov, line 1: Revisit this task.
It does not function as it should

Justsvetoslavov (changes requested):

PR #217, opened 2023-01-04
Pract.09: Add 01-06 | Radina Ovcharova

radi676:
Radina Ovcharova 
6 3MI0600151


SI/Sem.09/Pract.09/Task06.cpp
```diff
@@ -0,0 +1,24 @@
+﻿#include <iostream>
+const int MAX_SIZE = 20;
+void addElement(int* array, int toAdd, int len)
+{
+	if (len == 0 || toAdd > array[len - 1])
+	{
+		array[len] = toAdd;
```
Justsvetoslavov, line 7: We need to resize here.
You have the MAX_SIZE.
array[len] is INVALID

Justsvetoslavov (changes requested):

SI/Sem.09/Pract.09/Task06.cpp
radi676: void resize(int*& array, int& size) {
	int* newArray = new int[size * 2];
	for (int i = 0; i < size; i++) {
		newArray[i] = array[i];
	}
	size *= 2;
	delete[] array;
	array = newArray;
}

void addElement(int* array, int toAdd, int len)
{
	resize(array, len);
	if (len == 0 || toAdd > array[len - 1])
	{
		array[len] = toAdd;
		return;
	}
	array[len] = array[len - 1];
	addElement(array, toAdd, len - 1);
}

Justsvetoslavov: @AtanasHrist0v will take over for task06 with an example solution.

Justsvetoslavov: Link: https://pastebin.com/2b1b3sE1?fbclid=IwAR2vAJXMvd5n-I9lyYZN1ff9wGDraNjnKAJtTVYy7RFHAbX3qeuoLCs1J-c

Justsvetoslavov (approved):

PR #218, opened 2023-01-08
Pract.09: Add 05-07 | Kristian Petrov

Kristian2109:
Kristian Petrov
6 8MI0600201


Justsvetoslavov (changes requested):
Task05 and Task06 have solutions.
Check them and if you think there is a better solution put a separate PR for them.
Task08 is empty?

Put a separate PR for task07 as it is the only task to be reviewed currently.
Closing the PR.

PR #219, opened 2023-01-08
pract.03:Add 07|Nikolai Chalakov

NikolaiChalakov:
Nikolai Chalakov
2 0MI0600260


Justsvetoslavov (changes requested):
Reformat the code (alignment) and remove unnecessary empty lines + comments.
After that we will start review of the task

PR #220, opened 2023-01-08
Pract.09: Add 09-10 | Ivan Stoynev

Vacata12:
Ivan Stoynev
3 6MI0700164


Justsvetoslavov (changes requested):
Revisit the tasks and think what can be done better. (not everything is functionality)

